PR #8, opened 2015-04-30
Python 3 Updates

levisaya:
Ran python-modernizer, this fixed most of the Python 3
incompatibilities.

Had to manually rework the iteration in svg_regex and svg_transform;
.next doesn't exist in Python 3. Wrapped the builtin next function in a
partial for equivalent functionality.


oberstet:
looks fine. thanks for the work!

Issue #9, opened 2015-06-15
ValueError: Attempted relative import in non-package


SimonSteinberger:
Using relative import paths inside scour.py doesn't work when calling the script from outside the package directory:

ValueError: Attempted relative import in non-package

    from .svg_regex import svg_parser

Should be

    from svg_regex import svg_parser

The same for two other relative imports inside this file.

oberstet:
Youd shouldn't call the script, but use the installed `scour` binary (https://github.com/oberstet/scour/blob/master/setup.py#L49)

PR #12, opened 2015-11-15
Allow the Scour module to be used as a standalone program

Ede123:
This PR allows to use the Scour module as a "standalone program" (e.g. download, unzip and run) without the need to install the module itself and/or any third party modules.

It consists of two parts:
* Use a local copy of "six". We should not expect six to be available in the users Python environment. E.g. Inkscape comes with a minimal installation of Python that does neither include six nor allows the user to easily install it.
* Add a simple wrapper script around the Scour module that can be executed directly. The relative imports introduced in d1c66cc75bbe474b5a45a7cd45ed27207311d897 have broken the previous approach to achieve this with `if __name__ == '__main__': run()`

### DIFF
--- a/scour.py
+++ b/scour.py
@@ -1,0 +1,26 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+#  Scour module wrapper script
+#
+#  Copyright 2015 Eduard Braun
+#
+#  This file is part of Scour, http://www.codedread.com/scour/
+#
+#   Licensed under the Apache License, Version 2.0 (the "License");
+#   you may not use this file except in compliance with the License.
+#   You may obtain a copy of the License at
+#
+#       http://www.apache.org/licenses/LICENSE-2.0
+#
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+
+# This file is a wrapper script around the actual Scour module (scour.py in scour folder)
+# It allows execution of Scour without installation of the module.
+
+from scour import scour
+scour.run()
--- a/scour/scour.py
+++ b/scour/scour.py
@@ -3280,8 +3280,3 @@ def start(options, input, output):
 def run():
    options, (input, output) = parse_args()
    start(options, input, output)
-
-
-
-if __name__ == '__main__':
-   run()
--- a/scour/scour.py
+++ b/scour/scour.py
@@ -60,8 +60,7 @@ from .svg_regex import svg_parser
 from .svg_transform import svg_transform_parser
 import optparse
 from .yocto_css import parseCssString
-import six
-from six.moves import range
+from . import six
 
 # Python 2.3- did not have Decimal
 try:
--- a/scour/six.py
+++ b/scour/six.py
@@ -1,0 +1,868 @@
+"""Utilities for writing code that runs on Python 2 and 3"""
+
+# Copyright (c) 2010-2015 Benjamin Peterson
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in all
+# copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+
+from __future__ import absolute_import
+
+import functools
+import itertools
+import operator
+import sys
+import types
+
+__author__ = "Benjamin Peterson <benjamin@python.org>"
+__version__ = "1.10.0"
+
+
+# Useful for very coarse version differentiation.
+PY2 = sys.version_info[0] == 2
+PY3 = sys.version_info[0] == 3
+PY34 = sys.version_info[0:2] >= (3, 4)
+
+if PY3:
+    string_types = str,
+    integer_types = int,
+    class_types = type,
+    text_type = str
+    binary_type = bytes
+
+    MAXSIZE = sys.maxsize
+else:
+    string_types = basestring,
+    integer_types = (int, long)
+    class_types = (type, types.ClassType)
+    text_type = unicode
+    binary_type = str
+
+    if sys.platform.startswith("java"):
+        # Jython always uses 32 bits.
+        MAXSIZE = int((1 << 31) - 1)
+    else:
+        # It's possible to have sizeof(long) != sizeof(Py_ssize_t).
+        class X(object):
+
+            def __len__(self):
+                return 1 << 31
+        try:
+            len(X())
+        except OverflowError:
+            # 32-bit
+            MAXSIZE = int((1 << 31) - 1)
+        else:
+            # 64-bit
+            MAXSIZE = int((1 << 63) - 1)
+        del X
+
+
+def _add_doc(func, doc):
+    """Add documentation to a function."""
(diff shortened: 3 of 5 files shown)


oberstet:
No, copy-pasting an external dependency (six) into this module is an anti-pattern ..

Ede123:
I don't like it either, but I like the external dependency even less... Do you have a better idea?

Would you accept a minimum set of compatibility code to replace "six" altogether?
Granted, "six" is a small module, but even a small module seems a bit excessive for a module as "simple" as Scour.

oberstet:
As said, duplicating code from a well established external package is a no go.

Rgd the wrapper script: that's already there https://github.com/codedread/scour/blob/master/setup.py#L49

Ede123:
I wasn't suggesting duplicating. I just think that what we currently try to achieve with "six" can be done with much less code in a much easier way. However if you're in favor of an external dependency let's stick to it and see how things play out.

Regarding wrapper script: This script is only created when installing the module, right?
This part of the PR request was specifically for the case where one *does not* want to install the module (e.g. for development purposes or to run multiple versions of Scour side-by-side on the same system). As a matter of fact I did all my recent development using this wrapper and as #9 indicates this could come in handy for others, too.
Do you have any reason why we shouldn't add it?

oberstet:
> This script is only created when installing the module, right?

Yes. It's a feature of distutils.

For development purposes, you can start scour like `python -m scour.run`.
For multiple versions of scour, there is virtualenv.

> Do you have any reason why we shouldn't add it?

Yes: there is an established approach: distutils / console_scripts.

Ede123:
> For development purposes, you can start scour like python -m scour.run
I figured that out, too, but only after I searched the Internet on how to do it. It's not an obvious/intuitive solution.

It might well be that your suggested approach is established amongst Python professionals (which you seem to be), but it surely complicates usage of Scour for everybody else.

Personally I always prefer software that keeps it as simple as running a single script/executable and does not need any additional steps like setup / installation of external libraries, etc so it's readily available for everybody (without a deeper understanding of e.g. Python / distutils / console_scripts / pip / whatever)
Maybe we simply differ in opinion here, but maybe it's also something for you to reconsider.

oberstet:
> It might well be that your suggested approach is established

Yes, it is.

And "established" == "simple", because of expectations, info on the Web etc etc.

btw: you should be able to directly run the : `python scour/scour.py` also because of this

https://github.com/codedread/scour/blob/master/scour/scour.py#L3305

Ede123:
> btw: you should be able to directly run the : python scour/scour.py also because of this

No, that specific code is actually broken after d1c66cc75bbe474b5a45a7cd45ed27207311d897 as mentioned in my initial description for this PR.
The linked commit introduced relative imports and trying to relatively import from a script does not work (one first has to import the parent module from an absolute path).
As I didn't find an easy way around this I decided on the wrapper script.

oberstet:
Huh? The code is using absolute imports https://github.com/codedread/scour/blob/master/scour/scour.py#L52

oberstet:
@Ede123 I've pushed https://pypi.python.org/pypi/scour/0.31

Ede123:
> Huh? The code is using absolute imports

I think https://github.com/codedread/scour/blob/master/scour/scour.py#L59-L60 as well as https://github.com/codedread/scour/blob/master/scour/scour.py#L62 are relative imports, aren't they?

When trying to execute the module as a script those lines error out (with an error message stating that a relative import is not possible without the parent module being imported absolutely; I can give you the exact error message when I get home where I have access to a machine with python installed)

> I've pushed https://pypi.python.org/pypi/scour/0.31
Great!

Ede123:
@oberstet: This is the error that is thrown when trying to execute the module as a script:
```
  File "scour.py", line 59, in <module>
    from .svg_regex import svg_parser
SystemError: Parent module '' not loaded, cannot perform relative import
```

oberstet:
@Ede123 ah, right. could you retry with trunk pls? just pushed https://github.com/codedread/scour/commit/302f7f7477193165b2b1bdcd5997ee4e1acbf997

Ede123:
I'm afraid this did not do it:
```
> python scour.py --version
  File "<frozen importlib._bootstrap>", line 2218, in _find_and_load_unlocked
AttributeError: 'module' object has no attribute '__path__'

During handling of the above exception, another exception occurred:

  File "scour.py", line 59, in <module>
    from scour.svg_regex import svg_parser
  File "E:\Temp\Downloads\Scour\scour.git\scour\scour.py", line 59, in <module>
    from scour.svg_regex import svg_parser
ImportError: No module named 'scour.svg_regex'; 'scour' is not a package
```

oberstet:
sure. try: `python scour/scour.py --version`. anyway: I won't change this. a package is a package.

Ede123:
Exactly the same issue... Did you try that on your machine? If it works I assume it only works because you have the module installed?

oberstet:
works.

```console
oberstet@thinkpad-t430s:~$ ./python279/bin/virtualenv ~/foobar
New python executable in /home/oberstet/foobar/bin/python
Installing setuptools, pip...done.
oberstet@thinkpad-t430s:~$ source ~/foobar/bin/activate
(foobar)oberstet@thinkpad-t430s:~$ which python
/home/oberstet/foobar/bin/python
(foobar)oberstet@thinkpad-t430s:~$ python
Python 2.7.9 (default, Jan 18 2015, 21:40:29) 
[GCC 4.8.2] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import scour
  File "<stdin>", line 1, in <module>
ImportError: No module named scour
>>> 
(foobar)oberstet@thinkpad-t430s:~$ which scour
/usr/local/bin/scour
(foobar)oberstet@thinkpad-t430s:~$ cd scm/oberstet/scour/
(foobar)oberstet@thinkpad-t430s:~/scm/oberstet/scour$ git pull
Already up-to-date.
(foobar)oberstet@thinkpad-t430s:~/scm/oberstet/scour$ make test_version
PYTHONPATH=. python -m scour.scour --version
  File "/home/oberstet/python279/lib/python2.7/runpy.py", line 162, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
  File "/home/oberstet/python279/lib/python2.7/runpy.py", line 72, in _run_code
    exec code in run_globals
  File "/home/oberstet/scm/oberstet/scour/scour/scour.py", line 60, in <module>
    from scour.svg_transform import svg_transform_parser
  File "scour/svg_transform.py", line 63, in <module>
    from six.moves import range
ImportError: No module named six.moves
make: *** [test_version] Fehler 1
(foobar)oberstet@thinkpad-t430s:~/scm/oberstet/scour$ which pip
/home/oberstet/foobar/bin/pip
(foobar)oberstet@thinkpad-t430s:~/scm/oberstet/scour$ pip install six
You are using pip version 6.0.6, however version 7.1.2 is available.
You should consider upgrading via the 'pip install --upgrade pip' command.
Collecting six
  Downloading six-1.10.0-py2.py3-none-any.whl
Installing collected packages: six

Successfully installed six-1.10.0
(foobar)oberstet@thinkpad-t430s:~/scm/oberstet/scour$ make test_version
PYTHONPATH=. python -m scour.scour --version
0.31
(foobar)oberstet@thinkpad-t430s:~/scm/oberstet/scour$ make test_help
PYTHONPATH=. python -m scour.scour --help
scour 0.31
Copyright Jeff Schiller, Louis Simard, 2010
Usage: scour.py [-i input.svg] [-o output.svg] [OPTIONS]

If the input/output files are specified with a svgz extension, then compressed
SVG is assumed. If the input file is not specified, stdin is used. If the
output file is not specified,  stdout is used.

Options:
  --version                   show program's version number and exit
  -h, --help                  show this help message and exit
  --disable-simplify-colors   won't convert all colors to #RRGGBB format
  --disable-style-to-xml      won't convert styles into XML attributes
  --disable-group-collapsing  won't collapse <g> elements
  --create-groups             create <g> elements for runs of elements with
                              identical attributes
  --enable-id-stripping       remove all un-referenced ID attributes
  --enable-comment-stripping  remove all <!-- --> comments
  --shorten-ids               shorten all ID attributes to the least number of
                              letters possible
  --shorten-ids-prefix=SHORTEN_IDS_PREFIX
                              shorten all ID attributes with a custom prefix
  --disable-embed-rasters     won't embed rasters as base64-encoded data
  --keep-unreferenced-defs    won't remove elements within the defs container
                              that are unreferenced
  --keep-editor-data          won't remove Inkscape, Sodipodi or Adobe
                              Illustrator elements and attributes
  --remove-metadata           remove <metadata> elements (which may contain
                              license metadata etc.)
  --renderer-workaround       work around various renderer bugs (currently
                              only librsvg) (default)
  --no-renderer-workaround    do not work around various renderer bugs
                              (currently only librsvg)
  --strip-xml-prolog          won't output the <?xml ?> prolog
  --enable-viewboxing         changes document width/height to 100%/100% and
                              creates viewbox coordinates
  -p DIGITS, --set-precision=DIGITS
                              set number of significant digits (default: 5)
  -q, --quiet                 suppress non-error output
  --indent=INDENT_TYPE        indentation of the output: none, space, tab
                              (default: space)
  --nindent=INDENT_DEPTH      depth of the indentation, i.e. number of
                              spaces/tabs: (default: 1)
  --no-line-breaks            do not create line breaks in output (also
                              disables indentation; might be overriden by
                              xml:space="preserve")
  --strip-xml-space           strip the xml:space="preserve" attribute from
                              the root SVG element
  --protect-ids-noninkscape   Don't change IDs not ending with a digit
  --protect-ids-list=PROTECT_IDS_LIST
                              Don't change IDs given in a comma-separated list
  --protect-ids-prefix=PROTECT_IDS_PREFIX
                              Don't change IDs starting with the given prefix
(foobar)oberstet@thinkpad-t430s:~/scm/oberstet/scour$ 
```

PR #14, opened 2015-11-17
Remove unused XML namespace declarations

Ede123:
This PR allows Scour to remove unused XML namespace declarations.

### DIFF
--- a/scour/scour.py
+++ b/scour/scour.py
@@ -2877,6 +2877,10 @@ def scourString(in_string, options=None):
    global numBytesSavedInTransforms
    doc = xml.dom.minidom.parseString(in_string)
 
+   # remove <metadata> if the user wants to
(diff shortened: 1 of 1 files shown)


oberstet:
Cool, thanks! LGTM code wise. I assume you know what you are doing with the XML mangling;) and have tested it - hence merging.

Ede123:
Thanks!

If you should ever have any comments code wise (even if it's only minor things) please let me know. Since this is my first shot at Python I'm sure I won't always find the most elegant solution and feedback is always welcome.

I tested the code for some files I had at hand. Of course one can never consider all possibilities but I think it should be fine.
In that regard: I noticed http://bazaar.launchpad.net/~scouring/scour/trunk/files had some unittests. Maybe we should bring them back?

PR #24, opened 2015-12-06
Import unittests from old repository

### LINKED ISSUE #15: Import unittests from old repository
As discussed at [#14](14#issuecomment-157532422) it would be nice to import the unittests from the old repository (see http://bazaar.launchpad.net/~scouring/scour/trunk/files).

Some preliminary findings:
* The unittests are run by executing `testscour.py` (which makes use of `testcss.py` and the test files in the folder `unittests`)
* It should be pretty straightforward to bring them back to life. I was already able to run the script with minimal adjustments. Some tests fail to execute (possibly due to Python 3 incompatibilities), apart from most tests seem to work (and to pass).

Ede123:
This PR includes the unittests from the old Launchpad repository (fixes #15).
They were updated to work with Python 3 and adapted to the current version of Scour.

As of now all tests can be run without errors.
I also fixed most of the smaller issues that caused some tests to fail along the way.
Right now there are two tests remaining that actually fail for valid reasons (I'll file new issues for them).


Ede123:
Everything OK with this PR?

I have a fix for #26 in line as part of which I'd also like to add another unittest which is why I decided to delay it until this PR is merged instead of submitting a separate PR right away.

Ede123:
Merged by the Maker himself. ;-)
Thanks!

Issue #26, opened 2015-12-06
Non-ASCII characters not handled properly


Ede123:
Importing the old unittests (#24) uncovered an issue with non-ASCII characters

See `unittests/utf8.svg`
```SVG
<?xml version="1.0" encoding="UTF-8" standalone="no"?>
<svg xmlns:xlink="http://www.w3.org/1999/xlink" 
   xmlns="http://www.w3.org/2000/svg">
  <desc>ú</desc>
</svg>
```
for the failing example

PR #27, opened 2015-12-08
Fix character encoding issues

Ede123:
This PR fixes character encoding issues (#26) by
* always reading binary data (both from files and from `stdin`) and letting the XML parser deal with the decoding
* always writing binary data (both to files and to `stdout`) since we assume an UTF8 encoding for the output and already encode it accordingly

Tests on Windows 7 with Python 3.5.0 and Python 2.7.11 as well as on Ubuntu 15.04 with Python 2.7.9 worked well for files with UTF and ISO 8859-15 encoding. I tested both file input as well as `stdin`/`stdout` input/output.


oberstet:
awesome! thanks for that!

PR #28, opened 2015-12-10
Simplify and fix "removeComments()"

### LINKED ISSUE #25: Comments not properly stripped
Importing the old unittests (#24) uncovered an issue with stripping comments via `--enable-comment-stripping`

See `unittests/comment-beside-xml-decl.svg`
```SVG
<?xml version="1.0" encoding="utf-8" standalone="yes"?>
<!-- Oh look a comment -->
<!-- generated by foobar version 20120503 -->
<!-- And another -->
<svg xmlns="http://www.w3.org/2000/svg">
  <!-- This comment is meant to test whether removing a comment before <svg>
       messes up removing comments thereafter -->
  <!-- And this one is meant to test whether iteration works correctly in
       <svg> as well as the document element -->
</svg>
```
for the failing example

Ede123:
* The separate treatment of comments at the documentElement's level is not necessary - they have a parent (as tested in Python 3.5.0 and 2.7.11 and 2.6.6)! It might not have worked before due to a typo - note the "if isinstance(element,...)" and "len(element.data)" which should both refer to "subelement" instead - or a bug in very old versions of Python).
* Fix the iteration over childNodes (i.e. replace "for subelement in element.childNodes:" with ""for subelement in element.childNodes[:]:". We have to create a copy of the list to iterate over, otherwise we'd be iterating over a list as we change it which leads to unpredictable results.

Fixes #25

### DIFF
--- a/scour/scour.py
+++ b/scour/scour.py
@@ -2587,20 +2587,11 @@ def removeComments(element) :
    """
    global numCommentBytes
 
-   if isinstance(element, xml.dom.minidom.Document):
-      # must process the document object separately, because its
-      # documentElement's nodes have None as their parentNode
-      for subelement in element.childNodes:
-         if isinstance(element, xml.dom.minidom.Comment):
-            numCommentBytes += len(element.data)
-            element.documentElement.removeChild(subelement)
-         else:
-            removeComments(subelement)
-   elif isinstance(element, xml.dom.minidom.Comment):
+   if isinstance(element, xml.dom.minidom.Comment):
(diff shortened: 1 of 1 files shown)


oberstet:
Thanks in particular for the detailed description of "why?". Both of these are quite subtle issues ..

oberstet:
@Ede123 I've pushed https://pypi.python.org/pypi/scour/0.32

Ede123:
:+1:

Issue #41, opened 2016-01-14
Installation via setup.py fails when not called from within package directory


Ede123:
Steps to reproduce:
* Downloaded https://pypi.python.org/packages/source/s/scour/scour-0.32.tar.gz and extract it
* Go to any folder but the folder the above package was installed to (in my case it was the directory containing the python binary)
* Try to install Scour using `python /path/to/Scour/setup.py install` as explained in `readme.md`

Result:
```
Traceback (most recent call last):
  File "..\..\scour-0.32\setup.py", line 35, in <module>
    verstrline = open(VERSIONFILE, "rt").read()
IOError: [Errno 2] No such file or directory: 'scour/__init__.py'
```

Probable source of the issue:
https://github.com/codedread/scour/blob/master/setup.py#L34-L35
`VERSIONFILE ` uses a relative path, so it is not found when not executed from within the package directory.

Eitot:
I added a PR for this. I am not sure whether this is the right way to do it, but it seems to work in my tests.

oberstet:
the package can be installed using pip or from within the source folder. the latter can be used for developers, but end users should always use pip.

oberstet:
ah, the readme is wrong!

oberstet:
I have updated the README: https://github.com/codedread/scour/blob/master/README.md

Eitot:
The readme isn't wrong, there is an error in `setup.py`. It should make reference to its local path before referencing components relative to its path. This is what I addressed with my PR, which seems to work (even without scour installed, of course). I would appreciate if you could explain why you think that this does not work.

Why did you remove Homebrew and everything else?

Ede123:
Well, not really an improvement in my opinion either... Why do we have to force people to use `pip`?

For example I noticed this issue when I was adding Scour to a Python distribution that is intended to be bundled with an installer which does not provide `pip`. Naively asked: What do I do now?

Why don't we simply try something like `os.path.dirname(os.path.realpath(__file__))` or ` sys.argv[0]` to get the absolute path of the directory that contains `setup.py`?

Eitot:
@Ede123: I considered that option as well, but for local python modules this is not even necessary. That is what the `import scour from __version__` was for. It loads the variable from `scour/__init__.py`. The exact same thing is done in `scour.py` on line 79.

oberstet:
https://github.com/codedread/scour/commit/73ec7da13e668ae6555f20daa673c29659f390a4

Ede123:
[...]

PR #42, opened 2016-01-15
Fix import error in setup.py

Eitot:
Addresses #41

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -16,7 +16,7 @@
 ##
 ###############################################################################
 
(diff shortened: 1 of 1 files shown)


oberstet:
this is wrong: it only works when the package is already installed

Eitot:
I cannot confirm that. Import does what it says it does: it imports the `__version__` variable from `scour/__init__.py`, regardless whether scour is installed or not.

PR #43, opened 2016-01-16
Hopefully definite fix for building from outside package directory

Ede123:
(follow-up for 73ec7da13e668ae6555f20daa673c29659f390a4 and #41)

`find_packages` does not seem to work correctly when the working directory does not equal the package directory (resulted in the actual module not being installed).
Fix this by changing the working directory during setup.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -32,7 +32,11 @@ Authors:
   - Tobias Oberstein (maintainer)
 """
 
-VERSIONFILE = os.path.join(os.path.dirname(os.path.realpath(__file__)), "scour", "__init__.py")
+old_path = os.getcwd()
+src_path = os.path.dirname(os.path.realpath(__file__))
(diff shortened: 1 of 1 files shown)


oberstet:
No, this is even more hackish. Please use pip for installation.

Ede123:
Well, I [borrowed that code from numpy](https://github.com/numpy/numpy/blob/master/setup.py#L333]https://github.com/numpy/numpy/blob/master/setup.py#L333) and [even setuptools itself does it this way](https://bitbucket.org/pypa/setuptools/src/cf547c694dc5a4f51623e4c5f944322ef9093e16/setup.py?fileviewer=file-view-default#setup.py-12) and I don't see why it should be considered hackish but you'll surely have your reasons...

Eitot:
@Ede123: Just fork it and add it.

PR #44, opened 2016-02-19
Mechanism for sanitizing options

Ede123:
This PR adds an easy way to sanitize an `options` object that should be passed to `scourString`/`scourXmlFile`.
By generating a set of default options and "merging in" know attributes from the options passed to the function it is ensured that all attributes are correctly set (default values are used for unspecified attributes) while also removing unknown attributes.

Firstly this makes it easier to use Scour from another script (Instead of requiring explicit calling of `generateDefaultOptions` to get a copy of the latest options which have to be modified and fed back into Scour it's enough to generate an object that contains just the attributes that one wants to change).
Secondly it will avoid compatibility issues that may arise in case any options are added/removed/renamed in the future.

### DIFF
--- a/scour/scour.py
+++ b/scour/scour.py
@@ -2866,8 +2866,9 @@ def serializeXML(element, options, ind = 0, preserveWhitespace = False):
 # input is a string representation of the input XML
 # returns a string representation of the output XML
 def scourString(in_string, options=None):
-   if options is None:
-      options = _options_parser.get_default_values()
+   # sanitize options (take missing attributes from defaults, discard unknown attributes)
+   options = sanitizeOptions(options)
+
    getcontext().prec = options.digits
    global numAttrsRemoved
    global numStylePropsFixed
@@ -3287,6 +3288,18 @@ def generateDefaultOptions():
    d = parse_args(args = [], ignore_additional_args = True)[0].__dict__.copy()
 
    return Struct(**d)
+
+
+
+# sanitizes options by updating attributes in a set of defaults options while discarding unknown attributes
+def sanitizeOptions(options):
+   optionsDict = dict((key, getattr(options, key)) for key in dir(options) if not key.startswith('__'))
+
+   sanitizedOptions = _options_parser.get_default_values()
+   sanitizedOptions._update_careful(optionsDict)
+
+   return sanitizedOptions
+
 
 
 def start(options, input, output):
--- a/testscour.py
+++ b/testscour.py
@@ -45,30 +45,31 @@ def walkTree(elem, func):
 		if walkTree(child, func) == False: return False
 	return True
 
+
 class ScourOptions:
-	simple_colors = True
-	style_to_xml = True
-	group_collapse = True
-	group_create = False
-	strip_ids = False
-	strip_comments = False
-	shorten_ids = False
-	shorten_ids_prefix = ""
-	embed_rasters = True
-	keep_defs = False
-	keep_editor_data = False
-	remove_metadata = False
-	renderer_workaround = True
-	strip_xml_prolog = False
-	enable_viewboxing = False
-	digits = 5
-	indent_type = "space"
-	indent_depth = 1
-	newlines = True
-	strip_xml_space_attribute = False
-	protect_ids_noninkscape = False
-	protect_ids_list = None
-	protect_ids_prefix = None
+	pass
+
+
+class EmptyOptions(unittest.TestCase):
+	def runTest(self):
+		options = ScourOptions
+		try:
+			scour.scourXmlFile('unittests/ids-to-strip.svg', options)
+			fail = False
+		except:
+			fail = True
+		self.assertEqual(fail, False, 'Exception when calling Scour with empty options object')
+
+class InvalidOptions(unittest.TestCase):
+	def runTest(self):
+		options = ScourOptions
+		options.invalidOption = "invalid value"
+		try:
+			scour.scourXmlFile('unittests/ids-to-strip.svg', options)
+			fail = False
+		except:
+			fail = True
+		self.assertEqual(fail, False, 'Exception when calling Scour with invalid options')
 
 class NoInkscapeElements(unittest.TestCase):
 	def runTest(self):


oberstet:
Looks good and makes sense! Thanks for explaining the motivation behind this PR too!

Ede123:
Thanks for merging!

Since it's related to this PR and I noticed you're also using it in [Taschenmesser](https://github.com/oberstet/taschenmesser/blob/d6945f6dbe5d0a84d33be6256de752ce25d31049/taschenmesser/svg.py#L45):
Do you see a reason for the complicated design of the function [`generateDefaultOptions()`](https://github.com/codedread/scour/blob/7c3331d6f6596c396c1343350d5f39f8390f6242/scour/scour.py#L3282) (the comment even states this is a "mess/hack" and should be fixed)? Wouldn't it suffice to simply return `_options_parser.get_default_values()`?

oberstet:
@Ede123 Yep, I agree https://github.com/oberstet/taschenmesser/issues/4

Essentially, I think `generateDefaultOptions` should be killed. I think `sanitizeOptions()` is much better. If we make that an official interface, a PEP8 compliant name like `make_options(options)` might be more intuitive. So the public, Python level API of Scour would be

* `scour.start(options, instream, outstream)`
* `scour.make_options(options)`

Maybe we could make it even more convenient / intutive:

* `scour.start(instream, outstream, options=None)`: use default options if none are given
* `scour.make_options(options=None)`: make default options if no override dict is provided

---

Medium term, we could think about moving to `argparse` or even [click](http://click.pocoo.org/) for command line parsing.

The nice thing with above new `sanitizeOptions` (or renamed to `make_options`): does not care if we use OptionsParse, argparse or click for CLI parsing .. that's properly encapsulated away.

Ede123:
`sanitizeOptions()`  already works with `options=None` so renaming might be all that's needed.

Do you think we should keep `generateDefaultOptions` as an alias for backwards compatibility? There's one other project on GitHub that [uses this function](https://github.com/hchauvet/beampy/blob/HEAD/beampy/functions.py#L183), but I guess it's not very widespread otherwise.

---

Regarding `argparse` - I already had a look at it: It should be easy to switch so could be done at some point. On the other hand I doubt we'd gain much of it right know (at least not without putting some additional thought into argument handling), so I wouldn't assign it a high priority right now.

oberstet:
> alias for backwards compatibility

Given that `generateDefaultOptions` probably isn't used a lot and that it was never announced as public API, IMO just killing it is ok. Plus:

* ping above user
* major version bump (eg to 0.4)

We might also take the chance and import all the _public_ API of Scour (that is, the 2 functions above) to `scour/__init__.py` and it add it to `__all__` within that file.

That would allow Python aware tools to automatically discover the _public_ API.

User would then do:

```python
from scour import start, make_options
```

and that's it.

No more `from scour import scour; scour.start`  ...

Above is considered "good practice" in many Python projects. So Scour is getting a better, and better Python citizen;) Which is nice and welcome I'd say.

oberstet:
Rgd argparse: I agree, it's not high prio, and has risks and chances for trouble, and no real added value (without doing even more work, and introduce user visible changes)

Ede123:
> We might also take the chance and import all the public API of Scour (that is, the 2 functions above) to scour/__init__.py and it add it to __all__ within that file.
If we do that I'd additionally consider `scourString` (and potentially `scourXmlFile`) as public API. It seems `scourString` is by far the most popular way to use Scour amongst other applications.

oberstet:
Agreed. We should aim for an official public API that we then promise to support.

Probably worth thinking about consistent naming then. There seem to be 3 types of things one might want to put through Scour:

* strings (of strict type `unicode`)
* files (file-like objects which are read/written binary and decoded/encoded UTF8)
* objects of type `xml.dom.minidom`

There would be 2 ways of exposing such functions:

1. a single procedure `convert()` which dynamically checks it's argument types and operates accordingly.
2. 3 functions named like `convert_string`, `convert_file` and `convert_dom`

Ede123:
Option 1 might overcomplicate things. Correctly handling a multitude of input formats might prove challenging while at the same time it makes it unclear for the user what data types the function is supposed (or even able) to work on.

I'd therefore prefer separate functions with clearly defined input parameters.
This also leads us to a potential problem: We'd also need clearly defined output parameters.
Should we presume that input and output types always match? In that case `scourXmlFile` already violates this assumption.

Should we even go a step further, instantiate a Scour class and define methods like
`Scour.input_string() / Scour.input_file() / Scour.input_dom()`
`Scour.output_string() / Scour.output_file() / Scour.output_dom()`
so the user can pick as required? This would require some effort to implement though and might also be an unnecessary overcomplication at this time, could however prove to be useful in the long run.


Last but not least I want to be picky :wink:
If we're going to create a fixed API in the way you suggested I want the functions to include the verb "scour". Where not converting any SVGs, where "scouring" them. A little bit of brand affinity doesn't hurt and it's definitely clearer if the source code contains a call to `scour_string` as compared to  `convert_string` (which could mean almost everything).

oberstet:
prefix the function with "scour",  where scour is interpreted as a verb, ok, cool;)

rgd variants: maybe the output should not be an argument to the function at all, but the return value.


```python
with open('infile', 'rb') as input, open('outfile', 'wb') as output:
    output.write(scour.scour(input.read()))
```

just thinking loudly.

we should continue discussing variants a little before settling. cause in the end, whatever we decide, that should be sane and stable ..

Ede123:
Yes, that's basically what I meant. Here some pseudo-code of what I intended:
```python
# setup an instance (with initial or default options)
my_scour = Scour(options)

# feed the class input by using any of the available input functions (we could easily add new ones for all types of input)
my_scour.input_file(infile)

# create the output (again we could easily add new output formats as needed)
outstring = my_scour.output_string()
outdom = my_scour.output_dom()
my_scour.output_file(outfile)

...
```
Except for `output_file()` (where we'd obviously need to pass a file handle and/or file name as parameter) the function returns the scoured data structure.

oberstet:
Mmh, I'd prefer retaining a simple function based interface, not a class based one:

```python
import scour

with open('input.svg', 'r') as infile, open('output.svg', 'w') as outfile:
    scour.optimize(infile, outfile, options={'shorten_ids_prefix': 'foo'}))
```

So the public API would boil down to:

* `scour.optimize(infile, outfile, options=None) -> result` operating on File-like (unicode based) objects, returning a result plain dict (with attributes like size before/after, run-time, but also the effective options applied in converting) and taking an options plain dict to override defaults
* `scour.run()` CLI runner (called from `__main__`), parsing command line args, opening input/output files, calling `scour.optimize` and printing the statistics info

> I'd screw `xml.dom.minidom` and plain strings as a API input/output types. And I think `scour.optimize` is more explicit in what it does than `scour.scour`, and the latter doubles the marketing message "scour".

Scour internally should then use https://docs.python.org/2/library/xml.dom.minidom.html#xml.dom.minidom.parse, not parseString, which only creates a temporary StringIO object anyway.

oberstet:
Some notes ..

Rgd opening file like objects:

```console
oberstet@thinkpad-t430s:~$ python
Python 2.7.6 (default, Jun 22 2015, 17:58:13) 
[GCC 4.8.2] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> f = open('.profile', 'r', encoding='utf8')
  File "<stdin>", line 1, in <module>
TypeError: 'encoding' is an invalid keyword argument for this function
>>> from io import open
>>> f = open('.profile', 'r', encoding='utf8')
>>> s = f.read()
>>> type(s)
<type 'unicode'>
>>> 
oberstet@thinkpad-t430s:~$ python3
Python 3.4.3 (default, Oct 14 2015, 20:28:29) 
[GCC 4.8.4] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> f = open('.profile', 'r', encoding='utf8')
>>> s = f.read()
>>> type(s)
<class 'str'>
```

Rgd processing DOMs with Unicode stuff .. compare the following with Python 2:

```console
oberstet@thinkpad-t430s:~$ python3
Python 3.4.3 (default, Oct 14 2015, 20:28:29) 
[GCC 4.8.4] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from xml.dom.minidom import parseString
>>> s = u'föö'.encode('utf8')
>>> s
b'f\xc3\xb6\xc3\xb6'
>>> d = parseString(u'<a>föö</a>')
>>> n = d.firstChild.firstChild.data
>>> type(n)
<class 'str'>
>>> n
'föö'
>>> n.decode('utf8')
  File "<stdin>", line 1, in <module>
AttributeError: 'str' object has no attribute 'decode'
>>> n.encode('utf8')
b'f\xc3\xb6\xc3\xb6'
>>> d = parseString(b'<a>föö</a>')
  File "<stdin>", line 1
SyntaxError: bytes can only contain ASCII literal characters.
>>> d = parseString(b'<a></a>')
>>> d = parseString(b'<a>foo</a>')
>>> n = d.firstChild.firstChild.data
>>> type(n)
<class 'str'>
>>> n
'foo'
>>> d = parseString(u'<föö>hello</föö>')
>>> n = d.firstChild.firstChild.data
>>> type(n)
<class 'str'>
>>> n
'hello'
>>> n = d.firstChild.firstChild
>>> n
<DOM Text node "'hello'">
>>> n = d.firstChild
>>> n
<DOM Element: föö at 0x7f2accb55d58>
>>> type(n.nodeName)
<class 'str'>
>>> n.nodeName.encode('utf8')
b'f\xc3\xb6\xc3\xb6'
>>> 
```

Ede123:
I could basically live with your [suggestion above](#issuecomment-186566042), BUT regarding

> I'd screw xml.dom.minidom and plain strings as a API input/output types. And I think scour.optimize is more explicit in what it does than scour.scour, and the latter doubles the marketing message "scour".

* I agree `xml.dom.minidom` is not necessary. There exist many more popular alternatives like etree and lxml for XML parsing that wouldn't be compatible anyway. We could check if there exists a generic class to store an XML tree that would be supported by multiple libraries but I couldn't find one so far. In the end passing string representations is probably the most compatible way which brings me to the next point.
* We *have* to support plain strings as this is a popular way to use Scour from within other scripts as already noted earlier. Data is not always available as a file on disk... (one could argue that one could convert strings to streams and pass them to Scour but then usage would get more complicated for no good reason)<br>This could well be achieved by additionally supporting `scour.optimize(instring, outstring, options=None) -> result`, but obviously with all the downsides I [mentioned above](#issuecomment-186363532). With proper documentation this might be OK though and would also allow us to do things like `scour.optimize(infile, outstring)` and `scour.optimize(instring, outfile)`
* I totally agree with you that `scour.scour` is bad and `scour.optimize` would work for me. Just to be sure: We'd keep the current module structure to ensure backwards compatibility, right?

PR #45, opened 2016-02-19
Switch order of new/old size in console output

Ede123:
This is just a small cosmetic in the console output.

Since it already looks like a fraction, I guess we should order new/old size accordingly so it actually represents the displayed quotient of file sizes.


oberstet:
Agreed .. easier to grasp.

codedread:
Hopefully it doesn't break downstream scripts - but LGTM :)

Issue #31, opened 2015-12-10
Continuous integration


oberstet:
Now that the unit tests are ported over from LP and do "work", we should think about setting up proper automation using Travis and tox to run all of these for each and every commit / PR on all platforms (Pythons) we want to support.

That way, we can make best use of the unit tests, with no ongoing manual work or risk of forgetting to run the unit tests.

oberstet:
This is now there .. click on the Build logo here https://github.com/scour-project/scour/blob/master/README.md

PR #52, opened 2016-04-02
Basic implementation for Travis CI

Ede123:
Here's a very basic implementation to use Travis CI, since it is time to tackle #31 to prevent glitches like #51 in future.

In principle it already works as tested in [my fork](https://github.com/Ede123/scour). So far it installs Scour, checks the installation by calling `scour --version` and runs `testscour.py`.
For example the initial build of master [failed](https://travis-ci.org/Ede123/scour/jobs/120202729), but my dummy pull request made it [pass](https://travis-ci.org/Ede123/scour/builds/120206007).

This is not (yet) meant to be a final PR, but to start discussion on what should be implemented (I'm just getting started with CI myself and am learning as I go), so let me know what you think.
On the other hand it seems to me even this basic functionality is already a big help and we probably don't even need much more...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,24 @@
+language: python
+
+install:
+  - pip install tox
+
+env:
+  - TOX_ENV=py26
+  - TOX_ENV=py27
+  - TOX_ENV=py33
+  - TOX_ENV=py34
+# - TOX_ENV=py35
+  - TOX_ENV=pypy
+
+script:
+  - tox -e $TOX_ENV
+
+matrix:
+  fast_finish: true
+
+  include:
+    # https://github.com/travis-ci/travis-ci/issues/4794#issuecomment-143758799
+    - python: 3.5
+      env:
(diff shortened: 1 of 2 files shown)


oberstet:
CI is highly welcome!! Thanks for working on that.

FWIW, a common way to do CI in Python project nowerdays probably is using a combination of Tox and Travis. Eg see https://github.com/crossbario/txaio - maybe have a quick peek into that, as it is doing things "very clean" and follows Python best-practices generally in many ways.

Ede123:
I will have a look at it (given that you don't shoot it down until it looks exactly like you imagined... then you'd better implement it yourself).

Ede123:
OK, I updated the PR. Does basically the same thing as before but this time using tox.
Latest test run with those files is [here](https://travis-ci.org/Ede123/scour/builds/120307796).

Let me know what you think...

oberstet:
@Ede123 awesome! a good start.

PR #53, opened 2016-04-02
add option to check and warn or bail out on flowtext

oberstet:
This fixes https://github.com/codedread/scour/issues/49 in a pragmatic way

### DIFF
--- a/scour/scour.py
+++ b/scour/scour.py
@@ -3237,6 +3237,7 @@ _option_group_compatibility = optparse.OptionGroup(_options_parser, "SVG compati
 _option_group_compatibility.add_option("--error-on-flowtext",
    action="store_true", dest="error_on_flowtext", default=False,
    help="In case the input SVG uses flow text, bail out with error. Otherwise only warn. (default: False)")
+_options_parser.add_option_group(_option_group_compatibility)
 
 
 def maybe_gziped_file(filename, mode="r"):
--- a/scour/scour.py
+++ b/scour/scour.py
@@ -2880,6 +2880,14 @@ def scourString(in_string, options=None):
    global numBytesSavedInTransforms
    doc = xml.dom.minidom.parseString(in_string)
 
+   # determine number of flowRoot elements in input document
+   # flowRoot elements don't render at all on current browsers (04/2016)
+   cnt_flowText_el = len(doc.getElementsByTagName('flowRoot'))
+   if cnt_flowText_el:
+      print("SVG input document uses {} flow text elements, which won't render on browsers!".format(cnt_flowText_el))
+      if options.error_on_flowtext:
+         sys.exit(1)
+
    # remove <metadata> if the user wants to
    if options.remove_metadata:
       removeMetadataElements(doc)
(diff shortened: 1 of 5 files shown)


scour/scour.py
```diff
@@ -2880,6 +2880,14 @@ def scourString(in_string, options=None):
...
    doc = xml.dom.minidom.parseString(in_string)
 
+   # determine number of flowRoot elements in input document
+   # flowRoot elements don't render at all on current browsers (04/2016)
+   cnt_flowText_el = len(doc.getElementsByTagName('flowRoot'))
+   if cnt_flowText_el:
+      print("SVG input document uses {} flow text elements, which won't render on browsers!".format(cnt_flowText_el))
+      if options.error_on_flowtext:
+         sys.exit(1)
+
```
Ede123: Right now it does not really matter (as it seems we don't touch `flowRoot`s, even if they are empty) but in general I guess "compatibility checks" should be done somewhere at the end of `scourString()` as we most likely want to check Scours output for compatibility, not the input?

scour/scour.py
```diff
@@ -3225,6 +3233,10 @@ def format_usage(self, usage):
    help="don't remove IDs starting with the given prefix")
 _options_parser.add_option_group(_option_group_ids)
 
+_option_group_compatibility = optparse.OptionGroup(_options_parser, "SVG compatibility checks")
+_option_group_compatibility.add_option("--error-on-flowtext",
+   action="store_true", dest="error_on_flowtext", default=False,
+   help="In case the input SVG uses flow text, bail out with error. Otherwise only warn. (default: False)")
```
Ede123: I think you're missing an `_options_parser.add_option_group(_option_group_compatibility)` here...

Ede123:
Why don't you include your tests in `testscour.py`? This will make it easier to keep track of functionality in future.

Also unittests should be minimal examples. Your SVG test files include a lot of unnecessary Inkscape cruft that has nothing to do with the actual test making it harder to find the cause of an issue if something breaks in future.

scour/scour.py
oberstet: Ah, good catch! fixed.

scour/scour.py
oberstet: In general, I agree.

In this case, why do all the hard work processing/cleansing the SVG, when we bail out anyway.

oberstet:
> Also unittests should be minimal examples

No. The should be as close to real world as possible. And that (foten) means, the input file comes from Inkscape (saved as Inkscape SVG). Because designers just press save.

> Why don't you include your tests in testscour.py? 

That's a good point. I will fix that, though it'll require some refactoring .. scour doesn't use exceptions at all! And the unit test don't spawn the scour executable, but call directly into the Python module, so we can't test for exit code 1.

Ede123:
> No. The should be as close to real world as possible.
Well, I guess these are two different philosophies, both having advantages and disadvantages.
The most annoying problem with huge test files is that they often cause false positives when something totally unrelated changes (especially if the test conditions are not well defined) which is likely to cause unnecessary work.

> That's a good point. I will fix that, though it'll require some refactoring .. scour doesn't use exceptions at all! And the unit test don't spawn the scour executable, but call directly into the Python module, so we can't test for exit code 1.
Yes, I see the need for a solution here, too, to check things like using file streams as input arguments. As you wrote we mainly test functionality in `scourString()`, but not the actual call of `scour` on the command line.

oberstet:
@Ede123 alright. unit tests added. I have only added minimal code .. no big refactoring. Too much work. Whitespace/PEP8 is also annoying and we should clean that up some time ..

oberstet:
FWIW, the whitespace commit in above only fixes trailing spaces .. the file contains tabs (probably even a mixture of spaces and tabs).

Many Python projects use this rule: indent == 4 spaces (no tabs). But PEP8 has more to say about whitespace ..

Issue #55, opened 2016-04-03
standalone="no" breaks some images


haarp:
I'm using Scour to optimize UI elements in Firefox. On some files this appears to break.

Scour adds **standalone="no"** to the header. This makes the file unable to be opened in Viewnior. If used as a Firefox UI element, it also fails to render (curiously it works if opened by Firefox as a page)

I don't understand what standalone means in the context of SVGs, so I can't say if this is a valid bug.

The file in question:
[menu-check-black.svg.zip](https://github.com/codedread/scour/files/201363/menu-check-black.svg.zip)

Ede123:
The standalone attribute tells the XML parser if it should use an external DTD for validation only (if no DTD is given it has no meaning at all), see for example [this explanation](http://www.xmlplease.com/xml/xmlquotations/standalone).

* Since `standalone="no"` is the default value, you should report the issue to the developers of your SVG-Viewer.
* Regarding Firefox: Could you check if the file works if the `standalone` attribute omitted? In that case the same holds as in the bullet point above. Maybe (and only maybe) Firefox requires SVG files used in the UI to be "standalone", though.

As for Scour I don't see anything wrong with adding `standalone="no"`, but since it's the default anyway we should probably omit and/or remove it unless the input document specifies `standalone="yes"`.

haarp:
Hmm. Strange indeed.

But if standalone=no is the default anyway, omitting it would both solve this problem aswell as shave off a few more bytes. Why not go for it? :)

haarp:
I did some further testing.

Firefox: Does not mind standalone=no. The problems Firefox had were actually #56.
GIMP: No problems either
Viewnior: Fails to open, "Couldn't recognize the image file format"
Xfce's thumbnail generator: Fails to generate thumbnails.

Both of the latter two are based on GTK2, which in turn uses librsvg for SVGs. But shouldn't librsvg be able to handle this? An upgrade of librsvg from 2.40.13 to 2.40.15 did not help.

Any ideas?

Ede123:
I just created a PR which makes scour omit the `standalone` attribute whenever it is not explicitly set to "yes".

I think this is everything we can do from the Scour side to help avoid this issue.

Ede123:
#57 is merged, so I'll close this issue.

PR #57, opened 2016-06-12
Only include "standalone" attribute if it was explicitly set to "yes"

Ede123:
"No" is the default value anyway...

(Also avoids #55)


Eitot:
Is this repository still being maintained? Last release was in January, despite several pull requests and changes since then.

oberstet:
@Eitot @Ede123 I've pushed https://pypi.python.org/pypi/scour/0.34

PR #59, opened 2016-07-29
add --order-attributes option

dirk-thomas:
If a tag has many attributes it can be hard to visually find the attribute you are looking for. Especially since many similar tags might have a very different order of the same attributes.

Therefore this patch adds a new option `--order-attributes` which orders the attribute alphabetically. It still keeps the `id` attributes in the front (as it was already before).

### DIFF
--- a/scour/scour.py
+++ b/scour/scour.py
@@ -2792,7 +2792,11 @@ def serializeXML(element, options, ind = 0, preserveWhitespace = False):
 
    # now serialize the other attributes
    attrList = element.attributes
-   for num in range(attrList.length) :
+   attrIndices = range(attrList.length)
+   if options.order_attributes :
+      attrName2Index = dict([(attrList.item(i).nodeName, i) for i in attrIndices])
+      attrIndices = [attrName2Index[name] for name in sorted(attrName2Index.keys())]
+   for num in attrIndices :
       attr = attrList.item(num)
       if attr.nodeName == 'id' or attr.nodeName == 'xml:id': continue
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Thank for merging this quickly. Do you happen to know what it would take to make this option available when exporting "optimized svg" from Inkscape?

oberstet:
nope, no clue. we use scour from CI / command line on normal inkscape svgs. maybe @Ede123 knows ..

Ede123:
I'm currently maintaining the Inkscape Scour extension and can add the 
necessary code.

It's to late to include it in 0.92 though, since we're in string freeze. I 
can give instructions though how to manually upgrade it when I'm back from 
holiday if required.


However what about making this a default (maybe not an option at all)? I 
don't see a reason not to always sort attributes in a reasonable way. In 
that case I'd favor a little more sophisticated sort order though (e.g. ID 
first, then default attributes of the object like x/y/width/height, then 
color/stroke style, Css style, and only then the other attributes in 
alphabetical order.


Am 1. August 2016 8:24:08 vorm. schrieb Tobias Oberstein 
<notifications@github.com>:

dirk-thomas:
<aking it a default - maybe with a more sophisticated order - sounds good to me too. If we agree on what the order should be I would be happy to provide another PR.

Ede123:
@dirk-thomas I opened #60 to further discuss making this a default as well as possible ordering mechanisms. Feedback is welcome.

PR #63, opened 2016-08-15
Enable quiet mode when writing output to stdout

### LINKED ISSUE #47: Output to `stdout` throws error
When writing output to `stdout`, (i.e. `scour -i input.svg > output.svg`) an error is thrown:
```
  File "C:\Python3.5\lib\runpy.py", line 170, in _run_module_as_main
    "__main__", mod_spec)
  File "C:\Python3.5\lib\runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "E:\Temp\Downloads\Scour\scour.git\scour\scour.py", line 3360, in <module>
    run()
  File "E:\Temp\Downloads\Scour\scour.git\scour\scour.py", line 3356, in run
    start(options, input, output)
  File "E:\Temp\Downloads\Scour\scour.git\scour\scour.py", line 3349, in start
    sizediff))
ValueError: I/O operation on closed file.
```

Ede123:
If quiet mode is not enabled when writing SVG output to stdout, all informational text output ends up mixed into the SVG output (notably the status report at the end, e.g. *"Scour processed file "x.svg" in 21 ms: 5708/13196 bytes new/orig -> 43.3%"*)

This fixes #47.

### DIFF
--- a/scour/scour.py
+++ b/scour/scour.py
@@ -3306,6 +3306,9 @@ def parse_args(args=None, ignore_additional_args=False):
          outfile = sys.stdout.buffer
       except AttributeError:
          outfile = sys.stdout
+      # enable quiet mode when writing output to stdout
+      # otherwise informational text output would be mixed into the SVG output
+      options.quiet = True
 
    return options, [infile, outfile]
 


Ede123:
@oberstet What's your opinion on this fix? Is this the "correct" or at least a reasonable way to do it?

Only other solution I see is to write all informational text output to stderr by default so it would not interfere when redirecting/piping the SVG output from stdout. But is this proper style?

Ede123:
The more I think about it the better I like the idea of using `stderr` for statistics (at least when writing output to `stdout`) instead of arbitrarily activating quiet mode.

Should this be the default when writing output to a file, too, or should we keep logging to `stdout` in that case?

Ede123:
Dropped in favor of #67.

PR #65, opened 2016-08-16
Print usage information if no input file was specified (and no data is available from `stdin`)

### LINKED ISSUE #34: Call without arguments
When I just type ``scour``, the console will show the version number and copyright notice, but it won’t terminate the Python process and return to the command line automatically. Is this intentional? Normally, I would expect that a call without any arguments either returns nothing or ``--help``.

Pressing control + c will also give me this output:
```
File "/usr/local/Cellar/scour/0.32/libexec/bin/scour", line 8, in <module> load_entry_point('scour==0.32', 'console_scripts', 'scour')()
File "/usr/local/Cellar/scour/0.32/libexec/lib/python2.7/site-packages/scour/scour.py", line 3321, in run
  start(options, input, output)
File "/usr/local/Cellar/scour/0.32/libexec/lib/python2.7/site-packages/scour/scour.py", line 3295, in start
  in_string = input.read()
KeyboardInterrupt
```

Note, I also have this when I install Scour with pip.

Ede123:
This fixes #34.

### DIFF
--- a/scour/scour.py
+++ b/scour/scour.py
@@ -3293,7 +3293,7 @@ def generateDefaultOptions():
        def __init__(self, **entries):
            self.__dict__.update(entries)
 
-   d = parse_args(args = [], ignore_additional_args = True)[0].__dict__.copy()
+   d = parse_args(args = [], ignore_additional_args = True).__dict__.copy()
 
    return Struct(**d)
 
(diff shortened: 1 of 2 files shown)


Ede123:
@oberstet The actual fix was tiny (c8eb29f54cdb12b702a2065be124c852cc40742b), however it revealed that we always tried to determine an input/output which is rarely ever required when not running Scour from command line (e.g. when running tests).

I therefore did some refactoring to split option parsing from getting input/output.

It should be fine but it probably wouldn't hurt to double-check in case I overlooked something.

PR #100, opened 2016-08-29
Remove obsolete Psyco dependency

Eitot:
This module does not work with Python 2.7 and later, according to https://pypi.python.org/pypi/psyco/1.6 and http://psyco.sourceforge.net.


Ede123:
Sounds reasonable...
While we do not aim for actively removing Python 2.6 compatibility, this certainly won't do any harm and I doubt there were many people who ever installed Psyco anyway.

PR #102, opened 2016-08-29
Add options to remove descriptive elements

### LINKED ISSUE #38: Option to remove descriptive elements
Following up on #35, I found that `<title>` and `<desc>` (as well as `<metadata>`) are purely descriptive elements, according to the [SVG 1.1 specification](http://www.w3.org/TR/SVG/intro.html#TermDescriptiveElement). There already is an option to remove `<metadata>` (`--remove-metadata`), but not for `<title>` and `<desc>`.

What we could do:
1.  Add `<title>` and `<desc>` to the same option.
2.  Add a new option that removes them (either separately or together).

Any thoughts?

Eitot:
Closes #38. I’ve added three new options:
- `--remove-titles` (removes `<title>`)
- `--remove-descriptions` (removes `<desc>`)
- `--remove-descriptive-elements` (removes `<title>`, `<desc>` and `<metadata>`)

I’ve also added tests for each (9 in total) and modified the corresponding SVG files. I’m not that familiar with Python, so constructive comments are appreciated! I wasn’t sure about adding a type check to `removeDescriptiveElements()` (for the set type), but apparently that is not very Python’y?

### DIFF
--- a/scour/scour.py
+++ b/scour/scour.py
@@ -910,11 +910,25 @@ def removeNamespacedElements(node, namespaces):
          num += removeNamespacedElements(child, namespaces)
    return num
 
-def removeMetadataElements(doc):
+def removeDescriptiveElements(doc, options):
+   elementTypes = []
+   if options.remove_descriptive_elements:
+      elementTypes.extend(("title", "desc", "metadata"))
+   else:
+      if options.remove_titles:
+         elementTypes.append("title")
+      if options.remove_descriptions:
+         elementTypes.append("desc")
+      if options.remove_metadata:
+         elementTypes.append("metadata")
+   if not elementTypes:
+      return
+
    global numElemsRemoved
    num = 0
-   # clone the list, as the tag list is live from the DOM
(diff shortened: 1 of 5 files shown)


scour/scour.py
```diff
@@ -3059,9 +3066,27 @@ def scourString(in_string, options=None):
       else:
          print("WARNING: {}".format(errmsg), file = options.ensure_value("stdout", sys.stdout))
 
-   # remove <metadata> if the user wants to
+   # remove descriptive elements
+   descriptiveElements = set()
```
Ede123: This is unnecessarily complex...
Just pass `options` to the function `removeDescriptiveElements(doc, options)` and do the required checking there.

You can then skip all the safety checks around line 914.

Also if you check for `remove_descriptive_elements` first you can potentially skip the check for the others and also do not require `Set.update()` (which consequently allows you to use a simple list).

Eitot:
@Ede123: Better?

Ede123:
Yes, perfect. Thank you!

Issue #119, opened 2016-09-22
Add public API


z0u:
I use `scour` as a module in a web app. Previously, I was doing this:

```python
    opts = scour.parse_args(args=[...])[0]
    output = scour.scourString(svg, opts)
```

But that fails in version 0.35, because `parse_args` now returns a single object instead of a tuple. The fix is easy; however, it would be nice to have a public API that would remain stable for a given major version number.

Ede123:
Yes, this is definitely something we want to offer.

To be able to assess this better:
* What would you need/expect from a public API?
* What's your usage scenario that makes you call `parse_args()` directly? I'd considered this function more or less private since I suspected argument handling is something every Python program does on its own...

My thoughts on what a minimal API should offer at least:
* I guess something equivalent to `scourString()` is essential, probably we could also expose `scourXmlFile()`
* For generating an options object we currently have `generateDefaultOptions()` and `sanitizeOptions()`. I'd suggest to expose something in the API that's basically a merged version of those two (i.e. a function that can be called without arguments to generate default options or with an existing options object to sanitize them).

Ede123:
I just created #121. Does this sound reasonable to you guys?

Any complaints @oberstet? Seems these are the parts that are less well documented in the Python world, but the current state looks quite workable from a usability point of view.

z0u:
Hi @Ede123, thanks for that - yes, that would suit our purposes very well.

> What's your usage scenario that makes you call parse_args() directly?

Simply to get an `options` object; I had just mimicked the operation of `run`. I didn't know about `generateDefaultOptions`.

> What would you need/expect from a public API?

Your suggestions are fine, and suit us well. The only thing I'd add is it's sometimes nice to have a builder-like API - but don't go to the effort of implementing it on my behalf, because I don't need it:

```python
options = options().with_quiet(True)
```

oberstet:
Here is a proposal for a public API in Python 3 type annotated syntax:

```python
def options(options: Optional[dict]) -> dict:
    """
    Returns a dictionary including attributes for all Scour options, either set to
    (normalized) values from the input options, or set to default values.
    The dictionary is JSON serializable (it only contains simple, serializable types).
    """

def scour_string(doc: str, options: Optional[dict]) -> str
    """
    Clean the SVG given in the string and return the clean SVG. The options must be
    created from `options()`.
    """
```

ewerybody:
Oh Hi! Was there any news?

[...]

Issue #103, opened 2016-08-29
Split scour.py into multiple files


Eitot:
Perhaps it would be worthwhile to split the huge scour.py file into smaller files and make the project slightly more object-oriented. I personally find it a bit intimidating and messy in its current state.

Ede123:
This is certainly something we want in the long run, but while there are many open issues about actual bugs, time is certainly better spent fixing those...

Also bear in mind that such a restructuring has to be done properly with a huge amount of care. Otherwise it only distributes the admittedly cluttered code across cluttered files. Even if done properly it's prone to regressions while using up huge amount of time...

Eitot:
Didn’t mean to imply that it wouldn’t be a big undertaking. I was just curious whether this is on your radar. 😉

oberstet:
I agree with both of you;) 

a) the huge single file sucks
b) it is not a simple file split up, but should be done rethinking the whole code structuring into functions/classes

So more  a mid term goal?

PR #121, opened 2016-09-25
Prepare public API

Ede123:
This is meant to be a first step towards a public API (#119)

Functions that are intended to be exposed via this API are collected in `__init__.py`. The current list is
* `scour_string` and `scour_xml_file` for the actual scouring
* `options()` to create default options (or sanitize an arbitrary object that might have fitting attributes)
* `parse_args()` in case someone prefers to create options from a string of command line arguments

With this Scour usage is as simple as
```Python
import scour

options = scour_options()
options.quiet = True

result = scour. scour_string("...", options)
```

I also added a `__main__.py` so execution of scour would be simply `python -m scour` instead of the redundant `python -m scour.scour`.

These changes also add an abstraction layer that should allow us to rename / move functions and split up source files (#103) in future without breaking compatibility.


Ede123:
As a matter of fact I wanted to change a few things since this first draft (most notably I remembered our discussion at #44) so I should probably close this as outdated for now until I have time to revisit it. The discussion can happen at #119 for the time being.

Would you be fine with having options accept a dict as input? (Would be easy to implement - basically we do that already in `sanitizeOptions()`).

I personally prefer `options()` to return an object-like value that has settable attributes. For one it's closer to the actual code (we pass the "options" object to almost every function), so if `options()` returned a compatible object it would make it easier to write unit tests and would allow third-party authors to use some internal functions directly (if they wanted to). Also I never quite got the popularity of dicts in Python. While they are easily parseable into objects I feel like they are "just" containers. If options was a real object we could even add error checking and the like in future (e.g. prevent users from setting options that do not exist, check input ranges, ...)

oberstet:
Ah, right. I have to agree with what I said there;) signature should read:

```python
def options(options: Optional[dict]) -> dict:
    """
    Returns a dictionary including attributes for all Scour options, either set to
    (normalized) values from the input options, or set to default values.
    The dictionary is JSON serializable (it only contains simple, serializable types).
    """
```

Ede123:
> signature should read:

Doesn't really resolve my main concern: I think it shouldn't return a dict.
I can not do anything useful with a dict anywhere in Scour, so unless we rewrite all functions to accept a dict it unnecessarily limits the usefulness of the returned object. Also it adds an unnecessary abstraction layer in all future implementations.

oberstet:
a dict is required as input to `scour_string`

Ede123:
You mean in your suggested API implementation?

oberstet:
yes. my main argument is: for a _public_ API for a thing like Scour, the types used should be JSON serializable. that makes integration easy, for non-py languages (cli), or for remoting (a scour service remotely callable), etc

oberstet:
maybe I should said that: in general, I'd agree with you that a python lib should use python objects in the public API. thinking about, we should probably first define what we mean about "public API"? what use cases mainly. because I had remoting and use from CLI _also_ in mind .. if you leave that out, and we talk about "python app using scour as python lib" as the one and only scenario, then my perspective changes ..

Ede123:
I'll have to think about that... Right now I have a feeling requiring a dict might slightly simplify some use cases, which are not very likely to start with, while complicating general usage and the internal code and also might make future expansions more complicated.

If a dict representation of the options is actually required, we could always implement a "dict()" method for our options object, so users could just do
```Python
options = scour.options()
dict_representation = options.dict()
```
(This is one of the expansions which would not be possible if we require a dict).

Edit: Well, it's hard to know in what ways Scour might be used by external applications. For me the aim was to offer a implementation that makes a "python app using scour as python lib" as easy as possible and would allow to be adjusted to our needs in future without breaking backwards.compatibility.

CLI usage is a question I could not solve myself yet: On the one hand we offer CLI already, on the other hand third-party apps might want to implement their own. Do we want to expose parts of the CLI functionality already present in Scour? Or should we make that part completely internal?  
I guess without actual usage scenarios it will be hard to design these parts of the API. Then again I assume usage scenarios will develop around whatever API we create, so the most flexible approach might just be the one to choose...

oberstet:
in that case I am against changing the current state and skip a public API: it allows me to bypass to where I want anyways: `options.__dict__`, because options is a simple value carrying class. burying that deeper in whatever scour type classes: as a user (and we are, via taschenmesser, eg https://github.com/oberstet/taschenmesser/blob/master/taschenmesser/svg.py), I don't care about those. I don't care about "design by book". I want scour to get out of my way and just use string -> string and options dict.

> Do we want to expose parts of the CLI functionality already present in Scour?

no

> Or should we make that part completely internal?

yes

oberstet:
so a compromise could be: use a `Options` class as today, but one which allows roundtrips to JSON / simple dicts, but nothing complicated for Options. essentially, leave it a simple value carrying class, not an opaque complex action/behavior carrying one.

PR #131, opened 2017-02-08
Improve precision options for smaller output size

pborunda:
Add a seperate precision option for curve control points
(--set-c-precision) . Interpret precision digits = 0 as round to nearest
integer between -1 and 1. Add option to preserves all digits left of
decimal point regardless of set precision.

### DIFF
--- a/scour/scour.py
+++ b/scour/scour.py
@@ -2539,7 +2539,7 @@ def serializePath(pathObj, options):
     """
     # elliptical arc commands must have comma/wsp separating the coordinates
     # this fixes an issue outlined in Fix https://bugs.launchpad.net/scour/+bug/412754
-    return ''.join([cmd + scourCoordinates(data, options, (cmd == 'a')) for cmd, data in pathObj])
+    return ''.join([cmd + scourCoordinates(data, options, (cmd == 'a'), cmd) for cmd, data in pathObj])
 
 
 def serializeTransform(transformObj):
@@ -2554,7 +2554,7 @@ def serializeTransform(transformObj):
     )
 
 
-def scourCoordinates(data, options, forceCommaWsp=False):
+def scourCoordinates(data, options, forceCommaWsp=False, cmd=''):
     """
        Serializes coordinate data with some cleanups:
           - removes all trailing zeros after the decimal
@@ -2567,7 +2567,7 @@ def scourCoordinates(data, options, forceCommaWsp=False):
         c = 0
         previousCoord = ''
         for coord in data:
-            scouredCoord = scourUnitlessLength(coord, needsRendererWorkaround=options.renderer_workaround)
+            scouredCoord = scourUnitlessLength(coord, needsRendererWorkaround=options.renderer_workaround, isControlPoint=((cmd == 'c' and (c % 6) < 4) or (cmd == 's' and (c % 4) < 2)))
             # only need the comma if the current number starts with a digit
             # (numbers can start with - without needing a comma before)
             # or if forceCommaWsp is True
@@ -2613,7 +2613,7 @@ def scourLength(length):
     return scourUnitlessLength(length.value) + Unit.str(length.units)
 
 
-def scourUnitlessLength(length, needsRendererWorkaround=False):  # length is of a numeric type
+def scourUnitlessLength(length, needsRendererWorkaround=False, isControlPoint=False):  # length is of a numeric type
     """
     Scours the numeric part of a length only. Does not accept units.
 
@@ -2625,7 +2625,26 @@ def scourUnitlessLength(length, needsRendererWorkaround=False):  # length is of 
 
     # reduce numeric precision
     # plus() corresponds to the unary prefix plus operator and applies context precision and rounding
-    length = scouringContext.plus(length)
+    sContext = scouringContext;
+    roundNearZero = scouringRoundNearZero
+    if(isControlPoint):
+       sContext = scouringContextC;
+       roundNearZero = scouringRoundNearZeroC;
+
+    if(roundNearZero and length > -1 and length < 1):
+       length = getcontext().create_decimal(str(int(round(length))))
+    else:
+       if(scouringKeepIntPrecision):
+          length_as_int = int(round(length))
+          len_length_as_int = len(str(abs(length_as_int)))
+          saved_prec = sContext.prec;
+          if(sContext.prec < len_length_as_int):
+             # preserve all digits left of the decimal point
+             sContext.prec = len_length_as_int;
+          length = sContext.plus(length)
+          sContext.prec = saved_prec;
+       else:
+          length = sContext.plus(length)
 
     # remove trailing zeroes as we do not care for significance
     intLength = length.to_integral_value()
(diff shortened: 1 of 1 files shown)


Ede123:
Hi pborunda, thanks for your contribution!

Before I start reviewing the code in detail:
What is the basic idea behind those two options? Only from looking at the code I see what you're doing but even I have difficulties understanding when I would want to use those options. I'm afraid normal users will be completely in the dark.

Therefore some general questions:
* In a language that everybody understands: What advantage do I have when using those options?
* From a technical perspective: Which kind of SVG files will profit from those options and under what circumstances? Feel free to include some real-world benchmarks.
* What are there significant advantages compared to the `--precision` option as-is? I dislike the fact that your PR seems to add a lot of complexity and possible side effects that are hard to predict...

pborunda:
This PR is especially useful for svg optimization where image contains a lot of curves. This PR is also useful when the svg optimization can benefit from low precision like 1 or 2 without loss of integer precision.

I've noticed that curveto [C and S] path sections do not much change the appearance when the curve control point precision is greatly reduced. On the other hand, changing the (X,Y) end points or anchor points to the same degree causes noticeable distortion to the image.

"--set-precision 0" is just like "--set-precision 1" but allows for better conversion from curveto into lineto
So for example
c0.123 0.223 0.323 0.423 55.6 6.7

with precision 1 (wont collapse to lineto)
"c.1 .2 .3 .4 60 7" 

with precision 0 (will collapse to lineto)
c0 0 0 0 60 7

"--keep-int-precision" This is an option I've always wanted. I've noticed that doesn't improve file size to reduce precision on integers.
example: take 12345.67 and 345.25
with precision 1 results: 10000 and 300
with precision 3 results: 12300 and 345
with precision 5 results: 12345 and 345.25

with --keep-int-precision option
with precision 1 results: 12345 and 345
with precision 3 results: 12345 and 345
with precision 5 results: 12345 and 345.25


So using the popular "kiwi.svg" compression example, i get an extra 43% reduction no visible distortion
--set-precision 5 (default) [RESULT 1956 bytes]
--set-c-precision 0 --keep-int-precision --set-precision 3 [RESULT 1112 bytes]

(above results combined with --strip-xml-prolog  --strip-xml-space --remove-metadata --indent=none --enable-comment-stripping --enable-id-stripping --enable-viewboxing)

The only side effect i see is if the user sets the --set-precision 6, control points will still be at precision 5. The user must explicitly set the control point precision, if the --set-precision is above the default 5. Setting below default is not a problem.

pborunda:
I fixed the side effect described above.

Ede123:
All right, I finished my investigations and here's my thoughts on the PR:

* With my recent commits integer precision is now automatically maintained when it's reasonable (i.e. whenever a better precision can be achieved without increasing the number of bytes). In my tests this made the `--keep-int-precision` option obsolete.
Are there any border cases where you think it could still be useful? Otherwise I'd drop it in favor of code clarity and consistency. Also every option that we save will be one option less for users to worry about.
* I really like the results of the reduced control point precision. It brings the file size down nicely while influencing the visual appearance much less than changing the overall precision would.
The problem I see: It's already hard for users to predict the influence of the precision parameter. Adding yet another variable does not exactly make it easier. Do you have any ideas how to reasonably improve upon that (i.e. by setting control point precision relative to precision) to make this parameter useful for a wider audience?
* The "rounding near zero" part: It complicates the code a lot and savings were only a few bytes at best in the tests I performed. Also it's kinda arbitrary to round near zero - why not reduce precision a bit everywhere else and save even more? It's also something that will be impossible for users to understand. How could "zero precision" be useful at all? Why does it influence rounding behavior? Why only near zero? If you don't have any strong reasons to keep it I think it should go, too.

Let me know what you think about this and we can plan how to go forward from there!

pborunda:
Thanks, I will test out your change related to #132 tomorrow. Yes agreed, the control point precision can be made relative to the overall precision. Maybe a percentage of overall precision (--set-c-precision 50% or --set-c-precision 0.5) or a precision offset (--set-c-precision-offset -2). Or drop the negative (-) sign of the offset value since the control point precision will always be less that the overall precision (--set-c-precision-offset 2). Also agreed with the round near zero, its not direct on it real intent which is to actually to trying convert more curveto sections into lineto sections

Ede123:
> Yes agreed, the control point precision can be made relative to the overall precision. Maybe a percentage of overall precision (--set-c-precision 50% or --set-c-precision 0.5) or a precision offset (--set-c-precision-offset -2).

I'm sorry, I think I wasn't clear enough: My question was if there is some universally useful factor/offset/whatever that would make it unnecessary for the user to decide on any number, i.e. a simple boolean switch like `--reduce-controlpoint-precision`. I guess you have more experience with that and can judge whether this is feasible. If not I'm fine with having a second numerical option.

> Also agreed with the round near zero, it's not direct on its real intent which is actually to trying convert more curveto sections into lineto sections. I can rerequest this "round near zero" suggestion as a separate PR if i can give a stronger argument.

I  think it certainly should be a second option but I'm doubting it's overall usefulness. In the files I tested it sometimes changed one control point to 0 or 1 from intermediate values, but that was about it. It certainly didn't produce straight line segments in those test files. (And even if it would, they would not be converted to lineto commands as `cleanPath()` was already called at that point).
If you have real world examples where this could substantially reduce file size I'm happy to reconsider but from what I saw savings are marginal at best.

pborunda:
I don't know of a best precision for control points, but i consistently get good results with control point precision 2. But this might not work for everyone especially if curves are very close to one another, so precision 3 should be a safe default. Its probably best to let the user decide the lowest acceptable precision for the job. I generally start at higher control point precision like 4 and then and then gradually reduce.

Real world example for the "round near zero".  I did a google search for svg and borrowed these 2 png files with many curves, and converted them to svg with Inkscape.
ship
squarl

ship.svg original 23.6KB
squirrel.svg original 14KB

after running "scour --strip-xml-prolog  --strip-xml-space --remove-metadata --indent=none --enable-comment-stripping --enable-id-stripping --enable-viewboxing" on both


ship.min.svg 19.6KB
squirrel.min.svg 11.3KB
I will use this as my baseline for below

options "--set-c-precision 1 --set-precision 4 --keep-int-precision"
ship.c1p4k.min.svg 13.3KB (67.4% of ship.min.svg)
squirrel.c1p4k.min.svg 7.11KB (62.7% of ship.min.svg)

options "--set-c-precision 0 --set-precision 4 --keep-int-precision"
ship.c0p4k.min.svg 11.8KB(60.4% of ship.min.svg)
squirrel.c0p4k.min.svg 6.55KB  (57.8% of ship.min.svg)

So its about 1.5KB smaller for ship.svg and 0.56KB smaller for squirrel.svg, about 10 percent smaller than "--set-c-precision 1"

Ede123:
Ah, I see, so it seems to work nicely for traced SVG files (probably because they have a lot of nodes very close to each other)

Could you open a new issue to discuss "rounding near zero" and attach the SVG files there?

Then we can focus on "--set-c-precision" here,

pborunda:
Ok will do. Just want to add your correct about the curveto not getting converted to lineto because of the order of the operations. 
I overlooked that originally. I see that have to rerun the scour optimization twice on the same file to get the simplification. Also it seems more likely that the curveto [C X1 Y1 X2 Y2 X Y] sections will be converted to curvto simplfied [S X1 Y1 X Y] section instead of lineto. Here's a section to give you an idea. 1st pass left, 2nd pass right.
untitled

pborunda:
Just confirming that your recent commit makes the --keep-int-precision obsolete. You can only see differences when testing with svg with very large viewbox over 10000x10000.

Ede123:
Going forward can you squash/rebase your PR and split out the `--keep-int-precision` and "round near zero" parts?

I'd like to do some re-factoring, too, but it's probably easiest if I do that on top of your changes.

pborunda:
Ok done. not sure about the build failure on de5aa60, but it works fine on command line. Is it an encoding issue? not sure

Ede123:
The built issue is fine (i.e. has nothing to do with your commit).
It's caused by 0f6d9be4e241ea2e53ea3cc8576789ac6ab341a2 and seems to be dependent on the machine the CI job is run on (as it worked at least once before without failing).

Issue #21, opened 2015-12-05
Quoted font-family-names get destroyed


jazzynico:
(Original report in https://bugs.launchpad.net/scour/+bug/1514073)

Font-family-names like "'MyriadPro-Regular'" get converted to "&apos;MyriadPro-Regular&apos;" with HTML-Entitys (primary fonts with whitespaces, because single quotation marks are recommended for this by the W3C) if you save as Optimzed SVG. Tested on Windows 7 x64 and Ubuntu latest version.

=====
Inkscape 0.91 r13725
includes Scour 0.26+r220

Also reproduced on Xubuntu 15.10 with Scour 0.31 (as a standalone script).
Test file available in the original report (unfortunately GH doesn't support SVG attachments...).

Ede123:
Fixed in 57f93efc894b1ba4c30e42b549b0081185029c36
Keep an eye open for regressions, since where not escaping quotes ('/") in text nodes and attribute names anymore.

Issue #56, opened 2016-05-04
Scour URL-escapes inline styles


haarp:
Apparently SVGs can contain inline CSS. Firefox for example makes use of that. I don't know their purpose, but here they are.

Scour seems to escape these styles.

    "

becomes

    &quot;

That doesn't seem right. attached is an example file: [radio.svg.zip](https://github.com/codedread/scour/files/249261/radio.svg.zip)

Ede123:
Also causes #21 which is probably the most common manifestation of this bug.

PR #148, opened 2017-08-25
Fix regression with regard to quotes in attribute values

mfwitten:
The following commit removed the escaping of single quotes and double quotes, in order to fix issues #21 and #56:

&nbsp;&nbsp;&nbsp;&nbsp;**commit** 57f93efc894b1ba4c30e42b549b0081185029c36  
&nbsp;&nbsp;&nbsp;&nbsp;**Date:** Tue Aug 16 00:10:41 2016 +0200

However, this broke the quoting of an attribute value that has both a single quote and a double quote. This patch series re-introduces the escaping of quotes, but ultimately uses entities only when both a single quote and a double quote are present at the same time in an attribute value; whether single quotes or double quotes are escaped depends on which choice produces the shortest result. Unit tests are included.

I didn't find out about issue #21 or #56 until just now, so it might be worthwhile to include tests for those situations before proceeding.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@
 
 ---
 
-Scour is an SVG optimizer/cleaner that reduces the size of scalable vector graphics by optimizing structure and removing unnecessary data written in Python.
+Scour is an SVG optimizer/cleaner that reduces the size of scalable vector graphics by optimizing structure and removing unnecessary data; scour is written in Python.
 
 It can be used to create streamlined vector graphics suitable for web deployment, publishing/sharing or further processing.
 
@@ -16,11 +16,11 @@ The goal of Scour is to output a file that renderes identically at a fraction of
 Scour is open-source and licensed under [Apache License 2.0](https://github.com/codedread/scour/blob/master/LICENSE).
 
 Scour was originally developed by Jeff "codedread" Schiller and Louis Simard in in 2010.
-The project moved to GitLab in 2013 an is now maintained by Tobias "oberstet" Oberstein and Eduard "Ede_123" Braun.
+The project moved to GitLab in 2013, and then later to GitHub; it is now maintained by Tobias Oberstein ("oberstet") and Eduard Braun ("Ede123").
 
 ## Installation
 
-Scour requires [Python](https://www.python.org) 2.7 or 3.3+. Further, for installation, [pip](https://pip.pypa.io) should be used.
+Scour requires [Python](https://www.python.org) 2.7 or 3.3+; for installation, [pip](https://pip.pypa.io) should be used.
 
 To install the [latest release](https://pypi.python.org/pypi/scour) of Scour from PyPI:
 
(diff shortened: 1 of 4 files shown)


Ede123:
Thank you for this PR Michael, looks very promising overall!

First things first:<br/>
Would it be too much to ask to split this PR? I'd be happy to merge some changes right away (e.g. 0df46a6cbc84023fcc575ab7ed504cf0608a0801, 95b1df843f9013a51fa62171d695e187269486c7), but I have some minor questions and modifications for others which I'd like to resolve first.
If you prefer to keep the PR intact at this point I understand but I think it makes reviewing and adjusting the individual commits a lot harder for both of us...

Regarding the central change:<br/>
Could you clarify the issue the current behavior causes for you? When I implemented 57f93efc894b1ba4c30e42b549b0081185029c36 I thought about possibilities where entities in attribute names could be desired but I came to the conclusion that they're probably invalid in all typical cases (as stated in my commit message).

mfwitten:
I've split this pull request into a number of quasi-separate pull requests: #149, #150, #151, #152. For now, I've left out the changes to `README.md`, they can be handled another time.

They are "quasi-separate", because each one forms the foundation for the next, so as to make for fast-forward merges without the need to do any rebasing. Of course, if there is a need to make changes to one of them, then they'll have to be rebased, anyway; however, I presume #149, #150, and #151 to be non-controversial.

Ede123:
Well, "quasi-separate" wasn't exactly what I had in mind (not much advantage in that... Ideally they would not depend on each other)

As a result I already messed it up (seems GitHub doesn't do fast-forward merging) so they all need to be rebased now...

Ede123:
You can combine #150 and #151 if you want (those should be ready to merge), I'll have another look at #152 afterwards...

mfwitten:
Sorry for the confusion.

I see now that GitHub uses `--no-ff` by default, and that the only way to get a fast-forward merge from GitHub's interface is basically to use the "Rebase and merge", which I suspect is what you used.

I've closed #150, and incorporated its commits into #151.

I've also rebased #151 onto the results from merging pull request #149, and I've decided to go ahead and rebase #152 onto the new #151, because that makes more sense to me than pretending like it was written on top of some earlier ancestor.

**However**, I suspect that I didn't need to do all of this rebasing on my end, anyway; by default `git rebase` throws out empty commits (ones that make no changes), and thus you should have been able to run "Rebase and merge" on both #150 and #151 without trouble, unless GitHub does something silly like `--keep-empty`.

So, now, after running "Rebase and merge" on #151: If you like #152 as it stands, you should be able to run "Rebase and merge" on it without worrying about the stale ancestors.

Ede123:
> I see now that GitHub uses --no-ff by default, and that the only way to get a fast-forward merge from GitHub's interface is basically to use the "Rebase and merge", which I suspect is what you used.

Yes, I used it for #149 and it seems to always rebase even if it would not be needed (i.e. fast-forward would be possible).

> However, I suspect that I didn't need to do all of this rebasing on my end, anyway; by default git rebase throws out empty commits (ones that make no changes), and thus you should have been able to run "Rebase and merge" on both #150 and #151 without trouble, unless GitHub does something silly like --keep-empty.

GitHub certainly does *something* as I wasn't able to merge the commits at all anymore after I initially used "Rebase and merge" on #149.<br/>
However one can merge on the command line with a fast-forward merge (which is what I did for #151), which then did not make #152 umergeable (even before you rebased it again).

PR #150, opened 2017-08-25
scour.py: minor rearrangement for the sake of clarity

mfwitten:
There has been a minor rearrangement of the code that handles the children of the element being serialized: The relevant `if` statement has had its condition effectively negated and thus has also had its *consequent* and *alternative* swapped; now, there is a very short *consequent*, followed by a very long *alternative*, rather than a very long *consequent* followed by a very short *alternative*.


mfwitten:
This pull request has been subsumed by #151.

PR #152, opened 2017-08-25
Fix regression with regard to quotes in attribute values

mfwitten:
The following commit removed the escaping of single quotes and double quotes, in order to fix issues #21 and #56:

&nbsp;&nbsp;&nbsp;&nbsp;**commit** 57f93efc894b1ba4c30e42b549b0081185029c36  
&nbsp;&nbsp;&nbsp;&nbsp;**Date:** Tue Aug 16 00:10:41 2016 +0200

However, this broke the quoting of an attribute value that has both a single quote and a double quote (see below for an example). This patch series re-introduces the escaping of quotes, but ultimately uses entities only when both a single quote and a double quote are present at the same time in an attribute value; whether single quotes or double quotes are escaped depends on which choice produces the shortest result. Unit tests are included.

-------------

The following presents a concrete, real-world example of the problem that is being fixed by this pull request.

    $ file='https://upload.wikimedia.org/wikipedia/commons/archive/a/a5/20170825000948%21Example_of_an_iterative_DNS_resolver.svg'
    $ curl "$file" >a.svg
    $ inkscape a.svg

In `inkscape`, select the text object that reads:

    "Where's www.wikipedia.org?"

Convert it to a path object (`Path > Object to Path`, or press `Shift+Ctrl+C`); next, save the file as `b.svg`, making sure to choose as the file type `Optimized SVG (*.svg)`. Then:

    $ grep -n Where b.svg
    43:  <g aria-label='"Where's www.wikipedia.org?"'>

Clearly, line 43 is wrong; you can check whether it's well formed, using a tool from the [expat](http://expat.sourceforge.net/) project:

    $ xmlwf b.svg
    b.svg:43:24: not well-formed (invalid token)
    $ sed -n -e '43{p;q;}' b.svg | cut -c 24-
    's www.wikipedia.org?"'>

Indeed, try opening the `b.svg` in a browser; `firefox 55.0.1` produces the following output (it even identifies a better character for the start of the error):

    XML Parsing Error: not well-formed
    Location: file:///tmp/b.svg
    Line Number 43, Column 25:
    
      <g aria-label='"Where's www.wikipedia.org?"'>
    ------------------------^

The browser of `Chrome OS 59.0.3071.134` produces:

    This page contains the following errors:
    
    error on line 43 at column 25: attributes construct error

    Below is a rendering of the page up to the first error.

This pull request fixes the problem.

### DIFF
--- a/scour/scour.py
+++ b/scour/scour.py
@@ -3178,21 +3178,23 @@ def remapNamespacePrefix(node, oldprefix, newprefix):
         remapNamespacePrefix(child, oldprefix, newprefix)
 
 
-def makeWellFormed(str):
-    # Don't escape quotation marks for now as they are fine in text nodes
-    # as well as in attributes if used reciprocally
-    #     xml_ents = { '<':'&lt;', '>':'&gt;', '&':'&amp;', "'":'&apos;', '"':'&quot;'}
+def makeWellFormed(str, quote=''):
     xml_ents = {'<': '&lt;', '>': '&gt;', '&': '&amp;'}
-
-#  starr = []
-#  for c in str:
-#      if c in xml_ents:
-#          starr.append(xml_ents[c])
-#      else:
-#          starr.append(c)
-
-    # this list comprehension is short-form for the above for-loop:
+    if quote:
+        xml_ents[quote] = '&apos;' if (quote == "'") else "&quot;"
     return ''.join([xml_ents[c] if c in xml_ents else c for c in str])
+
+
+def chooseQuoteCharacter(str):
+    quotCount = str.count('"')
+    aposCount = str.count("'")
+    if quotCount > aposCount:
+        quote = "'"
+        hasEmbeddedQuote = aposCount
+    else:
+        quote = '"'
+        hasEmbeddedQuote = quotCount
+    return (quote, hasEmbeddedQuote)
 
 
 # hand-rolled serialization function that has the following benefits:
@@ -3239,12 +3241,11 @@ def serializeXML(element, options, ind=0, preserveWhitespace=False):
     attrIndices += [attrName2Index[name] for name in sorted(attrName2Index.keys())]
     for index in attrIndices:
         attr = attrList.item(index)
-        # if the attribute value contains a double-quote, use single-quotes
-        quot = '"'
-        if attr.nodeValue.find('"') != -1:
-            quot = "'"
 
-        attrValue = makeWellFormed(attr.nodeValue)
+        attrValue = attr.nodeValue
+        (quote, hasEmbeddedQuote) = chooseQuoteCharacter(attrValue)
+        attrValue = makeWellFormed(attrValue, quote if hasEmbeddedQuote else '')
+
         if attr.nodeName == 'style':
             # sort declarations
             attrValue = ';'.join([p for p in sorted(attrValue.split(';'))])
@@ -3258,7 +3259,7 @@ def serializeXML(element, options, ind=0, preserveWhitespace=False):
                 outParts.append('xmlns:')
             elif attr.namespaceURI == 'http://www.w3.org/1999/xlink':
                 outParts.append('xlink:')
-        outParts.extend([attr.localName, '=', quot, attrValue, quot])
+        outParts.extend([attr.localName, '=', quote, attrValue, quote])
 
         if attr.nodeName == 'xml:space':
             if attrValue == 'preserve':
--- a/testscour.py
+++ b/testscour.py
@@ -1818,6 +1818,17 @@ class HandleQuotesInAttributes(unittest.TestCase):
                         'Failed on attribute value with the same number of double quotes as single quotes')
 
 
+class PreserveQuotesInStyles(unittest.TestCase):
+
+    def runTest(self):
+        with open('unittests/quotes-in-styles.svg', "rb") as f:
+            output = scourString(f.read())
+        self.assertTrue('use[id="t"]' in output,
+                        'Failed to preserve quote characters in a style element')
+        self.assertTrue("'Times New Roman'" in output,
+                        'Failed to preserve quote characters in a style attribute')
+
+
 class DoNotStripCommentsOutsideOfRoot(unittest.TestCase):
 
     def runTest(self):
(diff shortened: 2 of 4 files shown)


scour/scour.py
```diff
@@ -3178,20 +3178,10 @@ def remapNamespacePrefix(node, oldprefix, newprefix):
         remapNamespacePrefix(child, oldprefix, newprefix)
 
 
-def makeWellFormed(str):
-    # Don't escape quotation marks for now as they are fine in text nodes
-    # as well as in attributes if used reciprocally
-    #     xml_ents = { '<':'&lt;', '>':'&gt;', '&':'&amp;', "'":'&apos;', '"':'&quot;'}
+def makeWellFormed(str, quote=''):
```
Ede123: Could we put the new replacement logic below (in `serializeXML()`) into this function?

That way we'd keep all related functionality in a single function, a simple boolean argument (e.g. `escape_quotes`) could control behavior for attribute values vs. text nodes (i.e. disable the functionality in the latter case).

The type of quotes required could then be returned by the function.

Let me know what you think...

Edit: Since this would require some additional changes, let's complete this PR first and think about that suggestion later.... (Your input is still welcome; Let me know if you're interested in working on it)

scour/scour.py
```diff
@@ -3239,12 +3229,24 @@ def serializeXML(element, options, ind=0, preserveWhitespace=False):
...
+
+        quot_count = 0
+        apos_count = 0
+
+        for c in attrValue:
+            if c == '"':
+                quot_count += 1
+            elif c == "'":
+                apos_count += 1
+
```
Ede123: Any reason for using a for-loop here?

Can't we simply use [str.count](https://docs.python.org/3/library/stdtypes.html#str.count)?
(If it's significantly faster I'm fine with it, but I don't want to decrease legibility if it's only some negligible speedup).

scour/scour.py
```diff
@@ -3239,12 +3229,24 @@ def serializeXML(element, options, ind=0, preserveWhitespace=False):
...
+        if quot_count > apos_count:
+            quote = "'"
+        else:
+            quote = '"'
```
Ede123: I'd prefer `quote = "'" if quot_count > apos_count else '"'`

scour/scour.py
```diff
@@ -3597,9 +3597,7 @@ def xmlnsUnused(prefix, namespace):
 
 
 # used mostly by unit tests
-# input is a filename
-# returns the minidom doc representation of the SVG
-def scourXmlFile(filename, options=None):
+def scourXmlFileAndReturnString(filename, options=None):
```
Ede123: As a matter of fact we'll probably remove `scourXmlFile()` in favour of some (yet to be determined) public API eventually.

Could you rewrite the tests to use `scourString` directly instead of adding yet another "legacy" function?

scour/scour.py
mfwitten: Interestingly, my initial code did indeed use 2 calls to `str.count`, but as a matter of principle, I just couldn't stand the thought of looping through the same data twice (serially, no less!), when one loop would be enough.

Oh! Were that Python a more expressive language, with a more sophisticated interpreter! It'd be glorious to specify 2 calls to `str.count`, knowing that some kind of JIT would determine at run-time either to collapse them into one loop or spin them off into separate, parallel threads. Alas, our tools are not yet so beautiful.

Thus, I advocate for the single loop; however, how about abstracting that loop out into a function:

    def countQuotes(str):
        quotCount = 0
        aposCount = 0
    
        for c in str:
            if c == '"':
                quotCount += 1
            elif c == "'":
                aposCount += 1
    
        return (quotCount, aposCount)
    
    
    def chooseQuoteCharacter(str):
        (quotCount, aposCount) = countQuotes(str)
        quote = "'" if quotCount > aposCount else '"'
        return (quote, quotCount or aposCount)

Then, the code becomes something like:

    attrValue = attr.nodeValue
    (quote, hasEmbeddedQuote) = chooseQuoteCharacter(attrValue)
    attrValue = makeWellFormed(attrValue, quote if hasEmbeddedQuote else '')

Ede123: Let's do it with `count`... Unless I measured something wrong it's a lot faster:

```
import timeit

setup = '''
attrValue = '\'\"Where\'s www.wikipedia.org?\"\''
'''

a = '''
quot_count = 0
apos_count = 0

for c in attrValue:
    if c == '"':
        quot_count += 1
    elif c == "'":
        apos_count += 1

quote = "'" if quot_count > apos_count else '"'
'''

b = '''
quot_count = 0
apos_count = 0

quot_count = attrValue.count("'")
apos_count = attrValue.count('"')

quote = "'" if quot_count > apos_count else '"'
'''

number = 1000000
print(timeit.timeit(stmt=a, setup=setup, number=number))
print(timeit.timeit(stmt=b, setup=setup, number=number))
```

scour/scour.py
mfwitten: Certainly, that can be done, but my initial inquiry into the matter has led me to the conclusion that the result will actually be messier, mainly due to the way the unit tests will have to be written.

I think the resulting ugliness stems from the fact that `makeWellFormed` has no business choosing how to serialize a quoted string; using either single quotes or double quotes makes no difference to whether the result will be well formed&mdash;the choice should be the purview of a higher-level of abstraction, namely `serializeXML`, which makes the choice based on the criterion that the serialization should be as short as possible.

Also, what if an even higher level of abstraction (*e.g.*, the user) wants to ensure that every quoted string is delimited by double quotes, regardless of the length criterion? As it is now, it would be a simple matter of just telling `makeWellFormed` that a double quote is going to be used, and thereby letting `makeWellFormed` produced a well-formed value based on that fact.

scour/scour.py
mfwitten: Well, OK. I can't argue with those numbers.

Ede123:
I assume this is ready to be merged? (A quick comment would be helpful in future as adding/modifying commits does not trigger a notification)

I finally got around to do some final tests. It's looking good so far so I'll merge this now. Thanks for all your effort! Future contributions would certainly be appreciated!

mfwitten:
Great! Thanks for a productive interaction; in my experience, it's not uncommon for patches to languish without response from anybody.

Next time, I'll be sure to poke a pull request sooner. I was afraid of being annoying, so I was going to wait until some time next week to do so.

PR #165, opened 2018-02-17
Rewrite redundant codepatterns introduced by py2 -> py3 conversion

nthykier:
The automated python2 -> python3 converter creates some suboptimal
code patterns in some cases, notably in its handling of dicts.

This commit handles the following cases:

  * "if x in list(y.keys()):" => "if x in y:"

    The original code is neuters the O(1) lookup effeciency of a dict
    by turning it into a list.  This occurs a O(n) in converting it to
    a list and then another O(n) for the lookup.  When done in a loop,
    this becomes O(n * m) rather than the optimal O(m).

  * "for x in list(y.keys()):" => "for x in y:" OR "for x in list(y):"

    A dict (y in these cases) operates as an iterator over keys in the
    dict by default.  This makes the entire "list(y.keys())" dance
    redundant _in most cases_.  In a some cases, scour modifies the
    dict while iterating over it and in those cases, we need a
    "list(y)" (but not a "y.keys()").

    The benefit of this differs between python2 and python3.  In
    python3, we basically "only" avoid function call.  In python2,
    y.keys() generates a list, so here we avoid generating a
    "throw-away list".

The test suite succeed both with "python testscour.py" and "python3
testscour.py" (used 2.7.14+ and 3.6.4 from Debian testing).

On a 341kB flame-graph generated by "nytprof" (a perl profiler), this
commit changes the runtimes of scour from the range 3.39s - 3.45s to
3.27s - 3.35s making it roughly 3% faster in this case (YMMV,
particularly with different input).  The timings were recorded using
the following command line:

  time PYTHONPATH=. python3 -m scour.scour --enable-id-stripping  \
     --shorten-ids --indent=none  --enable-comment-stripping
     -i input.svg -o output.svg

This was used 5 times with and 5 times without the patch picking the
worst and best time to define the range.  The runtime test was only
preformed on python3.

All changed lines where found with:
  grep -rE ' in list[(].*[.]keys[(][)][)]:'

### DIFF
--- a/scour/scour.py
+++ b/scour/scour.py
@@ -844,7 +844,7 @@ def unprotected_ids(doc, options):
         protect_ids_list = options.protect_ids_list.split(",")
     if options.protect_ids_prefix:
         protect_ids_prefixes = options.protect_ids_prefix.split(",")
-    for id in list(identifiedElements.keys()):
+    for id in list(identifiedElements):
         protected = False
         if options.protect_ids_noninkscape and not id[-1].isdigit():
             protected = True
@@ -868,7 +868,7 @@ def removeUnreferencedIDs(referencedIDs, identifiedElements):
     global _num_ids_removed
     keepTags = ['font']
     num = 0
-    for id in list(identifiedElements.keys()):
+    for id in identifiedElements:
         node = identifiedElements[id]
         if id not in referencedIDs and node.nodeName not in keepTags:
             node.removeAttribute('id')
@@ -1049,7 +1049,7 @@ def moveCommonAttributesToParentGroup(elem, referencedElements):
 
         distinctAttrs = []
         # loop through all current 'common' attributes
-        for name in list(commonAttrs.keys()):
+        for name in commonAttrs:
             # if this child doesn't match that attribute, schedule it for removal
             if child.getAttribute(name) != commonAttrs[name]:
                 distinctAttrs.append(name)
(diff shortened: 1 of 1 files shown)


nthykier:
The test failure appear to be flake8 having obtained new rules since the last run:

```
flake8 runtests: commands[0] | flake8 --max-line-length=119
./testscour.py:69:9: E722 do not use bare except'
./testscour.py:79:9: E722 do not use bare except'
./testscour.py:94:9: E722 do not use bare except'
./testscour.py:112:9: E722 do not use bare except'
./scour/scour.py:3208:5: E741 ambiguous variable name 'I'
./scour/scour.py:3212:13: E741 ambiguous variable name 'I'
./scour/scour.py:3214:13: E741 ambiguous variable name 'I'
./scour/scour.py:3215:9: E741 ambiguous variable name 'I'
./scour/scour.py:3622:9: E722 do not use bare except'
```

None of these appear to be introduced by this commit itself.

Ede123:
Will look at this in a bit. I want to make extra sure this does not cause any regressions as similar code caused issues that were extremely hard to find before (as you note: sometimes the order really *is* important)

nthykier:
I suspect the big wins are the "if" statements and they are definitely not order sensitive.  If you prefer, I can split that out in a separate patch and we can just pick that. :)

Ede123:
OK, I think this is good to merge. Thanks for working on optimizing the code, this is much appreciated!

nthykier:
Thanks for fast reviews. :)

PR #167, opened 2018-02-17
Optimize find referenced elements

nthykier:
The biggest win (at least in my test case) is the "findReferencedElements: Handle referencingProps separately", which avoids a large number of unnecessary string operations.

But there is also a large potential in "Avoid recomputing findReferencedElements in removeUnusedDefs" for any SVG that need a number of calls to "removeUnreferencedElements" or where "removeUnusedDefs" recurses a lot of times.  My particular test case does not seem to trigger any of that though (with a total of one call to removeUnreferencedElements)

### DIFF
--- a/scour/scour.py
+++ b/scour/scour.py
@@ -586,10 +586,9 @@ def findReferencedElements(node, ids=None):
         # we remove the hash mark from the beginning of the id
         id = href[1:]
         if id in ids:
-            ids[id][0] += 1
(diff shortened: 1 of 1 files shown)


Ede123:
Perfect! 6% speedup also confirmed in a class of files I'm currently working with.

nthykier:
Thanks for the review and merge. :)

PR #168, opened 2018-02-17
Additional optimizations

nthykier:
The primary commit here is "Attempt to reuse results from findReferencedElements better", which reduces runtime for me by about 6%.  It does so by attempting to avoid calling findReferenceElements as it is the most expensive function there is.  In my case, it reduced the number of calls from 8 to 6 on the root element.

The next two commits rewrite findReferencedElements to an "iterative" variant.  In itself, it does not give anything measurable.  However, with those applied, it is a lot easier to tell how many times we call findReferenceElements on the root element (as it is now just the number of calls to findReferenceElements).

The last patch simply inlines a loop to remove some call overhead in python.  It is not particular measurable, but it does remove the function from the top 3 in the "ncalls" column from profiling.  With this, none of the scour.py functions appear in the top 15 of "ncalls" on my sample svg.

### DIFF
--- a/scour/scour.py
+++ b/scour/scour.py
@@ -561,53 +561,55 @@ def findReferencedElements(node, ids=None):
     global referencingProps
     if ids is None:
         ids = {}
-    # TODO: input argument ids is clunky here (see below how it is called)
-    # GZ: alternative to passing dict, use **kwargs
 
-    # if this node is a style element, parse its text into CSS
-    if node.nodeName == 'style' and node.namespaceURI == NS['SVG']:
-        # one stretch of text, please! (we could use node.normalize(), but
-        # this actually modifies the node, and we don't want to keep
-        # whitespace around if there's any)
-        stylesheet = "".join([child.nodeValue for child in node.childNodes])
-        if stylesheet != '':
-            cssRules = parseCssString(stylesheet)
-            for rule in cssRules:
-                for propname in rule['properties']:
-                    propval = rule['properties'][propname]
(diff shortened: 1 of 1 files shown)


Ede123:
I only had a quick glance so far, but two things I do not particulalry like about this PR:
* a8e4d0e (Attempt to reuse results from findReferencedElements better):
  This seems to require a lot of assumptions on when the referencedIDs stay valid and when they have to be recalculated. While I immediately believe it can improve execution times I'm afraid it does not help overall code quality (e.g. less modular, harder to maintain, easier to introduce regressions in future, etc.).<br>I like the general idea of optimizing calls to findReferencedElements, but I'm not sure this is the correct approach.
* c182fab  (Rewrite findReferencedElements without recursion)
  8c9c625 (Move loop inside a function to reduce the number of calls):
  As far as I understand this refactoring does not result in any measurable performance increase?
  If that's the case I'm not seeing the advantage as I feel the code becomes a bit more complicated for no gain... When I initially started working on Scour I liked the recursive implementation of many functions as it was elegant and easy to understand at the same time (and I still do). Reducing the number of times a function is called does not seem to be a priority unless it actually improves performance (or am I missing something?).

nthykier:
Is the ordering of nodes in the map returned by findReferencedElements relevant?  If not, I think I leverage that to try a different way to the same optimization.

Ede123:
I'd have to investigate myself...

nthykier:
Ok, will have another look at this from a different angle.

PR #169, opened 2018-02-17
Remove an unnecessary loop

nthykier:
The unprotected_ids function returns all unprotected ids and
removeUnreferencedIDs removes all of them that does not appear in the
return value of findReferencedElements.

On closer observation it turns out that removeUnreferencedIDs cannot
cause nodes/IDs to become unprotected nor unreferenced (as it only
remove the "id" attribute, not the node).  With this in mind, we can
just remove the loop and save a call to all of these functions.

### DIFF
--- a/scour/scour.py
+++ b/scour/scour.py
@@ -3471,11 +3471,9 @@ def scourString(in_string, options=None):
                 _num_elements_removed += 1
 
     if options.strip_ids:
-        bContinueLooping = True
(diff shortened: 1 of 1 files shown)


Ede123:
Good catch!

I think that's a remnant from before 28cc0d59fbb209163fb079dd67d177d151c6b58f when the loop was still removing elements.

PR #171, opened 2018-03-10
Avoid O(n²) in removeDuplicateGradient

nthykier:
The original implementation of removeDuplicateGradient does O(n²)
search over all gradients to remove duplicates.  In images with many
gradients (such as [MediaWiki_logo_1.svg]), this becomes a significant
overhead as that logo has over 900 duplicated gradients.

We solve this by creating a key for each gradient based on the
attributes we use for duplication detection.  This key is generated
such that if two gradients have the same key, they are duplicates (for
our purpose) and the keys are different then the gradients are
guaranteed to be different as well.  With such a key, we can rely on a
dict to handle the duplication detection (which it does very well).

This change improves the runtime performance on [MediaWiki_logo_1.svg]
by about 25% (8m51s -> 1m56s on 5 runs).

Original:
    $ time for I in $(seq 1 5) ; do \
        PYTHONPATH=. python3 -m scour.scour MediaWiki_logo_1.svg out.svg ; \
      done
    Scour processed file "heavy.svg" in 105042 ms: 1582746/4989544 bytes new/orig -> 31.7%
    Scour processed file "heavy.svg" in 103412 ms: 1582746/4989544 bytes new/orig -> 31.7%
    Scour processed file "heavy.svg" in 105334 ms: 1582746/4989544 bytes new/orig -> 31.7%
    Scour processed file "heavy.svg" in 107902 ms: 1582746/4989544 bytes new/orig -> 31.7%
    Scour processed file "heavy.svg" in 108161 ms: 1582746/4989544 bytes new/orig -> 31.7%

    8m51.855s
    ...

Optimized:
    $ time for I in $(seq 1 5) ; do \
        PYTHONPATH=. python3 -m scour.scour MediaWiki_logo_1.svg out.svg ; \
      done
    Scour processed file "heavy.svg" in 21559 ms: 1582746/4989544 bytes new/orig -> 31.7%
    Scour processed file "heavy.svg" in 21936 ms: 1582746/4989544 bytes new/orig -> 31.7%
    Scour processed file "heavy.svg" in 21540 ms: 1582746/4989544 bytes new/orig -> 31.7%
    Scour processed file "heavy.svg" in 21518 ms: 1582746/4989544 bytes new/orig -> 31.7%
    Scour processed file "heavy.svg" in 21664 ms: 1582746/4989544 bytes new/orig -> 31.7%

    real    1m56.400s
    ...

[MediaWiki_logo_1.svg]: https://upload.wikimedia.org/wikipedia/commons/archive/5/54/20120822053933%21MediaWiki_logo_1.svg

### DIFF
--- a/scour/scour.py
+++ b/scour/scour.py
@@ -57,7 +57,7 @@ import sys
 import time
 import xml.dom.minidom
 from xml.dom import Node, NotFoundErr
-from collections import namedtuple
+from collections import namedtuple, defaultdict
 from decimal import Context, Decimal, InvalidOperation, getcontext
 
 import six
@@ -1355,6 +1355,32 @@ def collapseSinglyReferencedGradients(doc):
     return num
 
 
+def computeGradientBucketKey(grad):
+    # Compute a key (hashable opaque value; here a string) from each
+    # gradient such that "key(grad1) == key(grad2)" is the same as
+    # saying that grad1 is a duplicate of grad2.
+    gradBucketAttr = ['gradientUnits', 'spreadMethod', 'gradientTransform',
+                      'x1', 'y1', 'x2', 'y2', 'cx', 'cy', 'fx', 'fy', 'r']
+    gradStopBucketsAttr = ['offset', 'stop-color', 'stop-opacity', 'style']
+
+    # A linearGradient can never be a duplicate of a
+    # radialGradient (and vice versa)
+    subKeys = [grad.getAttribute(a) for a in gradBucketAttr]
+    subKeys.append(grad.getAttributeNS(NS['XLINK'], 'href'))
+    stops = grad.getElementsByTagName('stop')
+    if stops.length:
+        for i in range(stops.length):
+            stop = stops.item(i)
+            for attr in gradStopBucketsAttr:
+                stopKey = stop.getAttribute(attr)
+                subKeys.append(stopKey)
+
+    # Use a raw ASCII "record seperator" control character as it is
+    # not likely to be used in any of these values (without having to
+    # be escaped).
+    return "\x1e".join(subKeys)
+
+
 def removeDuplicateGradients(doc):
     global _num_elements_removed
     num = 0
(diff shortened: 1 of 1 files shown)


nthykier:
This replaces #170

scour/scour.py
```diff
@@ -1355,6 +1355,32 @@ def collapseSinglyReferencedGradients(doc):
...
+    # radialGradient (and vice versa)
+    subKeys = [grad.getAttribute(a) for a in gradBucketAttr]
+    subKeys.append(grad.getAttributeNS(NS['XLINK'], 'href'))
+    stops = grad.getElementsByTagName('stop')
+    if stops.length:
+        for i in range(stops.length):
+            stop = stops.item(i)
+            for attr in gradStopBucketsAttr:
+                stopKey = stop.getAttribute(attr)
+                subKeys.append(stopKey)
```
Ede123: Can we guarantee this code will always produce unique keys?

E.g. when considering different number of stops / different attributes set? Especially with default attributes we produce a lot of repeating `\x1e\x1e`.

scour/scour.py
```diff
@@ -1355,6 +1355,32 @@ def collapseSinglyReferencedGradients(doc):
...
+
+    # Use a raw ASCII "record seperator" control character as it is
```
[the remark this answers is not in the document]
nthykier: Thanks; fixed in d79d955fa2820f0541ecfd0f4eb288ac9c00abdd (let me know if you want a rebase/squash)

scour/scour.py
nthykier: The number of elements per gradient and per stop is always constant as the append-operation is not conditional. They only case, where there might be a problem is if "attribute is defined explicitly to be empty" is different from "attribute is not defined at all" (both of those seem to make getAttribute return the empty string).  However, the original code did not distinguish between those two states, so I assumed I could rely on that as well.

That said, it should be a simple matter to use a different separator between each stop and a sentential value for "empty"/"undefined" attributes.

nthykier:
Thanks :) 

Re: Performance on a large number of distinct gradients: No, I do not have a test case for that lying around.  Nonetheless, assuming I do not add any bugs in the changeset, the performance for those should improve similarly provided that the total number of gradients.  If you see a speed regression, it will be on SVGs with a low number of gradient - but then the total cost of removeDuplicateGradients will become negligible anyway (as there is simply not a lot of work to do).

The changeset is an optimal CPU time/memory trade-off from an algorithm complexity point of view.  If you double the number of gradients:

 * Original code: The runtime of removeDuplicateGradients basically quad-doubles while memory doubles (i.e. runtime is O(n²) and memory is O(n))

 * Patched code: The runtime and memory usage of removeDuplicateGradients basically doubles (i.e. both are O(n)).  The base amount of memory per gradient is slightly higher compared to the original code.

As I understand it, this memory trade-off is not a concern for scour at the moment, given we load the entire XML into memory.

Ede123:
Just tested it on unique gradients and it's indeed ~5 times faster in my testcase.
I asked because the original code bailed out early in many cases, so I was not sure wether the reduced complexity of your code weighed out the additional computational effort, but that does not seem to be a relevant factor for usual files.

Issue #156, opened 2017-12-17
KeyError: u'linearGradient3326' with option --keep-unreferenced-defs


JoKalliauer:
The comand:
> scour -i [Oxygen15.04.1-edit-undo.svg](https://upload.wikimedia.org/wikipedia/commons/8/8d/Oxygen15.04.1-edit-undo.svg) -o output.svg --keep-unreferenced-defs

with the following inputfile:
[https://commons.wikimedia.org/wiki/File:Oxygen15.04.1-edit-undo.svg](https://commons.wikimedia.org/wiki/File:Oxygen15.04.1-edit-undo.svg)

lead to following error message:
```
Traceback (most recent call last):
  File "C:\Python27\Scripts\scour-script.py", line 11, in <module>
    load_entry_point('scour==0.36', 'console_scripts', 'scour')()
  File "c:\python27\lib\site-packages\scour\scour.py", line 3920, in run
    start(options, input, output)
  File "c:\python27\lib\site-packages\scour\scour.py", line 3888, in start
    out_string = scourString(in_string, options).encode("UTF-8")
  File "c:\python27\lib\site-packages\scour\scour.py", line 3496, in scourString
    while removeDuplicateGradients(doc) > 0:
  File "c:\python27\lib\site-packages\scour\scour.py", line 1424, in removeDuplicateGradients
    for elem in referencedIDs[dup_id][1]:
KeyError: u'linearGradient3326'
```

I added the option `--keep-unreferenced-defs` to avoid bug #155

Here is a minimal working example:
[min.svg.txt](https://github.com/scour-project/scour/files/1565763/min.svg.txt)
```svg
<?xml version="1.0" encoding="UTF-8" standalone="no"?>
<!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.0//EN" "http://www.w3.org/TR/2001/REC-SVG-20010904/DTD/svg10.dtd">
<svg id="svg2606" width="128" height="128" version="1.0" xmlns="http://www.w3.org/2000/svg">
  <defs
     id="defs2608">
    <linearGradient
       id="linearGradient3342">
      <stop
         id="stop3344"
         offset="0"
         style="stop-color:#000000;stop-opacity:1;" />
      <stop
         id="stop3347"
         offset="1"
         style="stop-color:#000000;stop-opacity:0;" />
    </linearGradient>
    <linearGradient
       id="linearGradient3326">
      <stop
         id="stop3328"
         offset="0"
[...]

Issue #164, opened 2018-02-09
--enable-id-stripping aswell as --shorten-ids changes colour


JoKalliauer:
Processing [https://commons.wikimedia.org/wiki/File:Test_only.svg](https://commons.wikimedia.org/wiki/File:Test_only.svg) with
> $ scour -i [Test_only.svg](https://upload.wikimedia.org/wikipedia/commons/5/58/Test_only.svg) -o output.svg --enable-id-stripping --shorten-ids

[min.svg.txt](https://github.com/scour-project/scour/files/1712746/min.svg.txt)
```svg
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN" "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd">
<svg version="1.1" viewBox="0 0 31 31" xmlns="http://www.w3.org/2000/svg">
 <style type="text/css"><![CDATA[
.group .shape_under_group,
.shape_direct
{
   fill:             #ff0000;
}
.square
{
   fill:             #0000ff !important;
}
.group #square_id
{
   fill:             #0000ff;
}
.force_inherit
{
   fill:             inherit;
}
]]></style>
 <g class="group" transform="translate(-56,-17)">
  <rect id="square_id" class="shape_under_group" x="56" y="17" width="31" height="31" fill="#00f"/>
 </g>
</svg>

```
![min](https://user-images.githubusercontent.com/15687581/36055283-b2c9bfe4-0dfb-11e8-9bf6-9d70b067ae94.png)

leads to
[output.txt](https://github.com/scour-project/scour/files/1712747/output.txt)
```svg
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE svg  PUBLIC '-//W3C//DTD SVG 1.1//EN'  'http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd'>
<svg version="1.1" viewBox="0 0 31 31" xmlns="http://www.w3.org/2000/svg">
 <style type="text/css"><![CDATA[
.group .shape_under_group,
.shape_direct
{
   fill:             #ff0000;
}
.square
{
   fill:             #0000ff !important;
}
.group #square_id
{
   fill:             #0000ff;
}
.force_inherit
{
   fill:             inherit;
}
]]></style>
 <g class="group" transform="translate(-56,-17)">
[...]

PR #173, opened 2018-03-10
Gracefully handle unreferenced gradients with --keep-unreferenced-defs

nthykier:
Closes: #156

### DIFF
--- a/scour/scour.py
+++ b/scour/scour.py
@@ -1430,22 +1430,26 @@ def removeDuplicateGradients(doc):
             # for each element that referenced the gradient we are going to replace dup_id with master_id
             dup_id = dupGrad.getAttribute('id')
             funcIRI = re.compile('url\\([\'"]?#' + dup_id + '[\'"]?\\)')  # matches url(#a), url('#a') and url("#a")
-            for elem in referencedIDs[dup_id]:
-                # find out which attribute referenced the duplicate gradient
-                for attr in ['fill', 'stroke']:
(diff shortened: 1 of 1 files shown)


Ede123:
Looks fine!

Before I merge a more fundamental question: Is it fine to de-duplicate gradients if `--keep-unreferenced-defs` is used to start with?

I tend towards "yes" as I assume `--keep-unreferenced-defs` is mostly used for non-gradients and even then a single copy of the unreferenced gradient should be enough unless the user has some very uncommon use case for duplicated and unreferenced gradients, but it seems like a good time to discuss this.

nthykier:
I think I am the wrong person to ask whether de-duplication makes sense here. :) I have no use-case for either behaviour (I just noticed that I now knew scour well enough to provide a potential patch for #156).

The only thing I can think of is: Do people rely on --keep-unreferenced-defs so they can reference it from parts of the SVG that scour does not understand/check for references?  (Issue #164 is the closest example I can think of, but I know too little SVG to say whether that case is applicable to this problem).

Ede123:
Well, obviously people could create CSS that overrides properties of gradients as well, but considering we do not account for any style attribute on gradients they are unfortunately "doomed" either way...

Scour currently only supports inline CSS, so using it in a different way is likely to cause undefined/broken behavior.

I think it's safe to remove duplicate gradients for now unconditionally (I can only try to make an educated guess myself...) and is probably what the majority of users would expect.

PR #175, opened 2018-03-10
Use time.perf_counter when available

nthykier:
As of python3.3, the recommended method for obtaining a performance
timer is no longer "time.clock()" but "time.perf_counter()".  By using
time.perf_counter, we outsource the problem for finding the best
performance messaurement to python.  Plus, we also avoid inaccurate
timings if the system time is changed during the processing of the SVG
image on platforms where we used to use "time.time()".

For python2 and python3 before 3.3, we still have to use the original
timers.

### DIFF
--- a/scour/scour.py
+++ b/scour/scour.py
@@ -74,11 +74,10 @@ VER = __version__
 COPYRIGHT = u'Copyright Jeff Schiller, Louis Simard, 2010'
 
 
-# select the most precise walltime measurement function available on the platform
-if sys.platform.startswith('win'):
-    walltime = time.clock
-else:
-    walltime = time.time
+# the walltime measurement function, we will use for reporting
+# reporting how long it took to process a given SVG file.  For our
+# purposes, the time.time() function has sufficent accuracy.
+walltime = time.time
 
 
 NS = {'SVG':      'http://www.w3.org/2000/svg',


Ede123:
Do you see any reason not to simply use `time.time()` on all OSs in all supported versions of Python?

It's not as if we needed a high performance counter here (it's only used for informational user output), and even on Windows the resolution is ~16 ms which seems more than sufficient for real-world usage.

nthykier:
I am not strongly tied to any of the functions; I just noted the comment saying we wanted the best timing function available and remembered that python3.3 had a better option here.  If time.time() is good enough and you are not concerned that it is deprecated for this purpose, then time.time() is fine for me.

Beyond that, the only case where I can think off an actual issue is if the system clock changes while scour is running (but it is still a relatively minor issue that the timing is off for a minimization).

Ede123:
That comment dates back to a time where the returned value might have been used for actually measuring performance and maybe ancient versions of Python (<2.7) that performed even worse. As I think overhead of calling scour is too high for high precision measurements anyway I either split out the parts I want to compare and time them separately with "timeit" or I use a sufficiently complex testcase, so the precision of the timer and overhead are negligible.

I think `time.time()` is not deprecated? (Only `timce.clock()` is as far as I can see).

I agree we can accept the dangers involved in a system clock changing durin processing. ;-)

nthykier:
Ok, time.time() it is. :)

Rebased patch coming up

Issue #161, opened 2018-01-11
scour needs unnecessary spaces in d-attribute of path-tag


JoKalliauer:
Processing [Discriptionpage:Latin letter Kĸ (kra) RSVG-BUG.svg](https://commons.wikimedia.org/wiki/File:Latin_letter_K%C4%B8_(kra)_RSVG-BUG.svg) with 
>scour [Latin_letter_Kĸ_(kra)_RSVG-BUG.svg](https://upload.wikimedia.org/wikipedia/commons/7/70/Latin_letter_K%C4%B8_%28kra%29_RSVG-BUG.svg) output.svg

```svg
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE svg  PUBLIC '-//W3C//DTD SVG 1.1//EN'  'http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd'>
<svg viewBox="0 0 723.09 231" xmlns="http://www.w3.org/2000/svg">
[...]

PR #177, opened 2018-03-10
Update the Lexer to accept expected tokens from the parser

nthykier:
The SVG spec allows paths to omit whitespace in some cases and expect
parsers to gracefully handle this.  In particularly, the parsers must
greedly match as much of the token required as possible but stop as
soon the token no longer matches.  The latter bit is where the SVG
standard gets interesting.

An Elliptical command (i.e. A or a) will accept among other:
  ..., number, flag, flag, number, ...

Where flag is defined as "0" or "1" (exactly one character).  Given
those tokens and the following input:

   1 010

The spec requires scour to parse that as:

  "1" (number), "0" (flag), "1" (flag) and "0" (number).

It might be tempting to just include "flag" in the default
tokenization.  Unfortunately this falls apart pretty quickly if you
want to follow the spec.  E.g. if you have 100 as input and the lexer
has no hint about the next token then it can now parse it as:

 * Three flags
 * Two flags and a one-digit number (in that order)
 * A flag and a two-digit number (in that order)
 * A three-digit number

Therefore, to support this, the SVGPathParser must provide the Lexer
with a hint about what it is expecting in some cases.  This turns out
to be trivially possible by exploiting the fact that "lex" is a
generator function and can trivially be converted to a "coroutine" (by
replacing "next(x)" with "x.send(value)").

### DIFF
--- a/scour/svg_regex.py
+++ b/scour/svg_regex.py
@@ -45,8 +45,6 @@ from __future__ import absolute_import
 
 import re
 from decimal import Decimal, getcontext
-from functools import partial
-
 
 # Sentinel.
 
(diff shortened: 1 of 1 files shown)


nthykier:
This would close #161

Ede123:
As mentioned in #161 I had already fixed this locally when you created the PR.

I compared both solutions:
* I actually consider mine a bit "hackish" but it's quite short and simple and does the job.
* Your code tries a more general approach and therefore requires a lot more code changes, however I feel that in the current form it's not really "better" than my approach - you still have to add a manual exception for boolean flags.

For these reasons I decided to use my code (see #183) but thanks for looking into this anyway!

Long term it would be great if we could have a path parser which we can actually teach the proper path grammar, so it automatically knows what to parse and will be able to detect errors in the path data, but for now I guess we're fine as-is.

nthykier:
Eduard Braun:
> As mentioned in #161 I had already fixed this locally when you created the PR.
> 

Indeed, I noticed after creating the pull request. :)  It happens.

> I compared both solutions:
> * I actually consider mine a bit "hackish" but it's quite short and simple and does the job.
> * Your code tries a more general approach and therefore requires a lot more code changes, however I feel that in the current form it's not really "better" than my approach - you still have to add a manual exception for boolean flags.
> 
> For these reasons I decided to use my code (see #183) but thanks for looking into this anyway!
> 

No problem. I am glad it is now fixed. :)

Thanks for considering,
~Niels

Issue #163, opened 2018-01-28
path processiong wrong


JoKalliauer:
Processiong [File:Envelope_string_art.svg](https://commons.wikimedia.org/wiki/File:Envelope_string_art.svg) with
> $ scour [Envelope_string_art.svg](https://upload.wikimedia.org/wikipedia/commons/6/6a/Envelope_string_art.svg) output.svg

connects a corner to the origin, instead of closing the rectangle:
[input.svg.txt](https://github.com/scour-project/scour/files/1671066/input.svg.txt)
```svg
<?xml version="1.0" encoding="UTF-8"?>
<svg width="600" height="480" xmlns="http://www.w3.org/2000/svg">
 <path d="m108.5 408h441.3m-441.3 0v-371.9m0 0v371.9h441.3v-371.9z" fill="none" stroke="#000"/>
</svg>

```
![input](https://user-images.githubusercontent.com/15687581/35482143-3edbfa58-0431-11e8-8aaa-122581419e50.png)

The output is:
[output.svg.txt](https://github.com/scour-project/scour/files/1671067/output.svg.txt)
```svg
<?xml version="1.0" encoding="UTF-8"?>
<svg width="600" height="480" xmlns="http://www.w3.org/2000/svg">
 <path d="m108.5 408h441.3m-441.3 0v-371.9 371.9h441.3v-371.9z" fill="none" stroke="#000"/>
</svg>

```
![output](https://user-images.githubusercontent.com/15687581/35482149-530c595a-0431-11e8-90ac-7e11bd7402e4.png)

so
`d="m108.5 408h441.3m-441.3 0v-371.9m0 0v371.9h441.3v-371.9z"` is changed to:
`d="m108.5 408h441.3m-441.3 0v-371.9 371.9h441.3v-371.9z"`

nthykier:
I have traced this issue to the following part of cleanPath:

```python
                if cmd == 'm':
                    # remove m0,0 segments
                    if pathIndex > 0 and data[0] == data[i + 1] == 0:
                        # 'm0,0 x,y' can be replaces with 'lx,y',
                        # except the first m which is a required absolute moveto
                        path[pathIndex] = ('l', data[2:])
                        _num_path_segments_removed += 1
                    else:  # else skip move coordinate
                        i = 2
```

The bug disappears if we special case on "len(data) == 2" and insist on keeping the "m0,0" command in that case.  However, that breaks trips the following test case:

```python
class PathEmptyMove(unittest.TestCase):

[...]

PR #178, opened 2018-03-11
Correct handling of "m0 0" vs. "z" commands

nthykier:
As observed in #163, scour sometimes do some unsafe path rewrites.

This branch solves that issue by separating the logic for removing path segments from the logic for rewriting commands.  The first commit removes the current unsafe optimization and the next one reimplements it (and adds a few more test cases for path cleaning of commands).

### DIFF
--- a/scour/scour.py
+++ b/scour/scour.py
@@ -2190,14 +2190,15 @@ def cleanPath(element, options):
             i = 0
             if cmd in ['m', 'l', 't']:
                 if cmd == 'm':
-                    # remove m0,0 segments
-                    if pathIndex > 0 and data[0] == data[i + 1] == 0:
-                        # 'm0,0 x,y' can be replaces with 'lx,y',
-                        # except the first m which is a required absolute moveto
-                        path[pathIndex] = ('l', data[2:])
-                        _num_path_segments_removed += 1
-                    else:  # else skip move coordinate
-                        i = 2
+                    # It might be tempting to rewrite "m0 0 ..." into
+                    # "l..." here.  However, this is an unsound
+                    # optimization in general as "m0 0 ... z" is
+                    # different from "l...z".
+                    #
+                    # To do such a rewrite, we need to understand the
+                    # full subpath, so for now just leave the first
+                    # two coordinates of "m" alone.
+                    i = 2
                 while i < len(data):
                     if data[i] == data[i + 1] == 0:
                         del data[i:i + 2]
--- a/scour/scour.py
+++ b/scour/scour.py
@@ -2181,10 +2181,11 @@ def cleanPath(element, options):
             x, y = startx, starty
             path[pathIndex] = ('z', data)
 
-    # remove empty segments
+    # remove empty segments and redundant commands
     # Reuse the data structure 'path' and the coordinate lists, even if we're
     # deleting items, because these deletions are relatively cheap.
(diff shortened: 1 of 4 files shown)


oberstet:
@nthykier I had a quick look at the code, but obviously, I would need to dig deep into actual SVG specification to verify;) However, obviously it addresses a real issue. So maybe we could "reduce the risk" of this change (because it touches deep internals) by adding more test cases/files? That would make me feel comfortable merging. @Ede123 ? I think adding more test cases / files, in _both_ synthetic and real-world would be very good also in general ..

Ede123:
I looked into this before and it looks mostly fine, although I wanted to double-check again before merging. It's certainly on the short-list for the next release.

Code-wise it's seems fine, although the new logic following `# remove no-op commands` is a bit hard to dig through - it might be an indication we can do even better - it might just be the spec that makes it unnecessarily complicated.

oberstet:
@Ede123 sounds good! 

> note: I have split out my "idea" above of adding a set of real world docs in CI as a new issue https://github.com/scour-project/scour/issues/194 for discussion

nthykier:
Eduard Braun:
> I looked into this before and it looks mostly fine, although I wanted to double-check again before merging. It's certainly on the short-list for the next release.
> 
> Code-wise it's seems fine, although the new logic following `# remove no-op commands` is a bit hard to dig through - it might be an indication we can do even better - it might just be the spec that makes it unnecessarily complicated.
> 

I am open to suggestions for improving the flow.  Please let me know if
there is anything concrete that makes it difficult for you to follow the
flow of the code.


The optimizations at a high level:

 * We can only remove "z" if the "pen" is at the same position as the
   start of the path (base case "m0 0z", but theoretically you can have
   steps in between "mX Y" and "z" as long as "z").

 * We can only rewrite m0 0 X Y to lX Y if 1) there is no "z" following
   the "m", OR 2) there is an "m" between this "m" and the "z".
   - This is because "z" is "relative" to the latest "m" before it.
   - Note that you can have multiple "z" cmds after the same "m"
     (a la: "m0 0 10 10-10-10 z l 10-10-10 10 z")

 * We want to remove redundant "m0 0" at the end of the path (e.g.
   because we optimized out a trailing "z").

In my code, I have used a "reverse iteration" (i.e. start at the end of
the path and move towards the start).  This has the advantage that we
can optimize path nodes immediately when we see them (assuming the
conditions are right).
  In a forward iteration, you find a node and have to look forward
because you can determine whether you can delete it.

Thanks,
~Niels

nthykier:
Hi @Ede123,

Would you prefer if we simply disabled the removal of "z"-segments (i.e. the reduce this pull request to the first commit only)?  This would ensure that scour does not corrupt the images while avoid the code flow that you are not happy with.

oberstet:
@nthykier just a quick note (to this and your other PRs): your work is welcome and much appreciated! I just find the time currently to dive into and review. we'll get it review and merged;)

Ede123:
OK, finally got around to finish my review and we're good to go (sorry for the delay...).

I'm still not sure wether there's a better way to do this, but your version works @nthykier, so let's go with it and thank you very much for all the effort again! As @oberstet already noted it's much appreciated!

nthykier:
Thanks for the review and merge. :)

Issue #129, opened 2017-01-19
Support in-place scrubbing


simon04:
Similar to `sed`. From its man page:
```
       -i[SUFFIX], --in-place[=SUFFIX]
              edit files in place (makes backup if SUFFIX supplied)
```

PR #180, opened 2018-03-18
Support bulk processing of SVG files

nthykier:
This changeset enables scour to process multiple files via the command line interface, which in turn can be used to reduce the overhead of calling scour by about 40%.

Adding bulk processing requires some interface changes:
 
- Multiple input files are accepted (if `-i` is used, it must be used for all input files)
- The output "file" can now be a directory and it must be a directory when multiple input files are specified)
- scourString no longer relies on options.infilename determining the "base directory" when embedding raster images. Instead, callers are expected to provide that directory when relevant.
- Because we no longer have a single inputfile, options.infilename is renamed to options.infilenames

### DIFF
--- a/scour/scour.py
+++ b/scour/scour.py
@@ -3017,7 +3017,7 @@ def removeComments(element):
     return num
 
 
-def embedRasters(element, options):
+def embedRasters(element, references_relative_to, options):
     import base64
     """
       Converts raster references to inline images.
@@ -3056,9 +3056,9 @@ def embedRasters(element, options):
             # relative local paths are relative to the input file, therefore temporarily change the working dir
             working_dir_old = None
             if parsed_href.scheme == 'file' and parsed_href.path[0] != '/':
-                if options.infilename:
+                if references_relative_to:
+                    working_dir_new = os.path.abspath(references_relative_to)
                     working_dir_old = os.getcwd()
-                    working_dir_new = os.path.abspath(os.path.dirname(options.infilename))
                     os.chdir(working_dir_new)
 
             # open/download the file
@@ -3308,7 +3308,7 @@ def serializeXML(element, options, indent_depth=0, preserveWhitespace=False):
 # this is the main method
 # input is a string representation of the input XML
 # returns a string representation of the output XML
-def scourString(in_string, options=None):
+def scourString(in_string, options=None, references_relative_to=None):
     # sanitize options (take missing attributes from defaults, discard unknown attributes)
     options = sanitizeOptions(options)
 
@@ -3549,7 +3549,7 @@ def scourString(in_string, options=None):
     # convert rasters references to base64-encoded strings
     if options.embed_rasters:
         for elem in doc.documentElement.getElementsByTagName('image'):
-            embedRasters(elem, options)
+            embedRasters(elem, references_relative_to, options)
 
     # properly size the SVG document (ideally width/height should be 100% with a viewBox)
     if options.enable_viewboxing:
@@ -3590,16 +3590,14 @@ def scourString(in_string, options=None):
 # used mostly by unit tests
 # input is a filename
 # returns the minidom doc representation of the SVG
-def scourXmlFile(filename, options=None):
+def scourXmlFile(filename, options=None, references_relative_to=None):
     # sanitize options (take missing attributes from defaults, discard unknown attributes)
     options = sanitizeOptions(options)
-    # we need to make sure infilename is set correctly (otherwise relative references in the SVG won't work)
-    options.ensure_value("infilename", filename)
 
     # open the file and scour it
     with open(filename, "rb") as f:
         in_string = f.read()
-    out_string = scourString(in_string, options)
+    out_string = scourString(in_string, options, references_relative_to)
 
     # prepare the output xml.dom.minidom object
     doc = xml.dom.minidom.parseString(out_string.encode('utf-8'))
@@ -3814,9 +3812,11 @@ def maybe_gziped_file(filename, mode="r"):
 
 
 def getInOut(options):
+    references_relative_to = None
     if options.infilename:
         infile = maybe_gziped_file(options.infilename, "rb")
         # GZ: could catch a raised IOError here and report
+        references_relative_to = os.path.dirname(options.infilename)
     else:
         # GZ: could sniff for gzip compression here
         #
@@ -3840,7 +3840,7 @@ def getInOut(options):
         # redirect informational output to stderr when SVG is output to stdout
         options.stdout = sys.stderr
 
-    return [infile, outfile]
+    return [infile, references_relative_to, outfile]
 
 
 def getReport():
@@ -3862,7 +3862,7 @@ def getReport():
     )
 
 
-def start(options, input, output):
+def start(options, input, output, references_relative_to=None):
     # sanitize options (take missing attributes from defaults, discard unknown attributes)
     options = sanitizeOptions(options)
 
@@ -3870,7 +3870,7 @@ def start(options, input, output):
 
     # do the work
     in_string = input.read()
-    out_string = scourString(in_string, options).encode("UTF-8")
+    out_string = scourString(in_string, options, references_relative_to).encode("UTF-8")
     output.write(out_string)
 
     # Close input and output files (but do not attempt to close stdin/stdout!)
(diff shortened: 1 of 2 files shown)


Ede123:
This is a welcome change but I think this needs some discussion:

#### Why is it slow in the first place?
Have you determined what causes the overhead when calling Scour for each file individually?  
* Is there anything slow in Scour or is it simply calling Python?
* Are the 40% overhead you mentioned for real-world data or for "empty" SVG files? (i.e. does the overhead matter when operating on typical files?

If something is slow in Scour I'd prefer to optimize that first

#### How to specify output file names
I'm not sure requiring a single output directory is the best way to approach this as it causes an asymmetry between processing single files and processing multiple files and would make it impossible to give precise file names to the output files. In your particular implementation it's also unclear what input files are and what the output file is (e.g. what would happen if the user types `scour in1.svg in2.svg in3.svg`? Or for that matter only `scour in1.svg in2.svg`?)

I'm not sure what the best option is myself yet, so any input is welcome.

Some possible implementations I can think of right now:
* Always expect groups of input/output files, i.e.  
  `scour -i in1.svg -o out1.svg -i in2.svg -o out2.svg`  
  which has the downside of requiring a lot of typing but would keep behavior consistent otherwise.
* Add separate options (e.g. `--outdir` and/or `--suffix`) that switch Scour into "batch processing mode" and deactivate the need for an output file name treating all unnamed parameters as input file names.  
  This could also work well alongside an `--inplace` option that would then overwrite input files (#129).

I'm also thinking about what should happen if a user calls scour like `scour *.svg`.

nthykier:
Eduard Braun:
> This is a welcome change but I think this needs some discussion:
> 

Ok :)

> #### Why is it slow in the first place?
> Have you determined what causes the overhead when calling Scour for each file individually?  

I have not.  However, my concrete usage for scour was to process the
auto-generated SVG graphs from gnuplot on a static generated website.
That site contains about ~1000 SVG files, even "small" things add up here.



> * Is there anything slow in Scour or is it simply calling Python?

It would seem that python is a significant part of it (although not the
only "slowness").  On my machine, an empty program (via -c '') costs
about 0.04s / 40ms.  For 1000 SVG files that becomes 40 seconds

"""
$ time python3 -c ''

real    0m0.046s
user    0m0.036s
sys     0m0.011s
"""

On the performance case I mentioned in my commit, python3's start up
time would account for roughly 4s out of 13s (~1/3) of the overhead.

> * Are the 40% overhead you mentioned for real-world data or for "empty" SVG files? (i.e. does the overhead matter when operating on typical files?
> 

Before I started submitting patches, the SVG files would take about
300ms a piece according to scour's own timing (i.e. excluding start up
time).  With the patches, I believe I got them down to about ~200ms a
piece in average.

What is a bit tricky is the timing in my patch is based on images that
were already scour'ed once (while the 300ms just above is not).

> If something is slow in Scour I'd prefer to optimize that first
> 

That is a good idea too.  However, for my use case, a bulk-processing
interface is going to be a significant improvement.

> #### How to specify output file names
> I'm not sure requiring a single output directory is the best way to approach this as it causes an asymmetry between processing single files and processing multiple files and would make it impossible to give precise file names to the output files.

I am happy to change the interface as long as it is compatible with
xargs.  For that, I basically need a static option in the beginning
(e.g. -o <dir>) and then be able to append "infinitely many input files
after that".

> In your particular implementation it's also unclear what input files are and what the output file is (e.g. what would happen if the user types `scour in1.svg in2.svg in3.svg`? Or for that matter only `scour in1.svg in2.svg`?)
> 

With the interface proposed then:

 * If you use -o, then that is always the output as always.
 * If there are multiple input files, then the output must be a
   directory.  There should be several checks to ensure this.

I would be perfectly fine with requiring -o for the output with multiple
input files if that would solve this concern for you.


> Some possible implementations I can think of right now:
> * Always expect groups of input/output files, i.e.  
>   `scour -i in1.svg -o out1.svg -i in2.svg -o out2.svg`  
>   which has the downside of requiring a lot of typing but would keep behavior consistent otherwise.

It would also make bulk processing via "xargs" or "find -exec ... +"
more difficult.  Those are the particular tools I am targeting.

> * Add separate options (e.g. `--outdir` and/or `--suffix`) that switch Scour into "batch processing mode" and deactivate the need for an output file name treating all unnamed parameters as input file names.  
>   This could also work well alongside an `--inplace` option that would then overwrite input files (#129).
> 
Indeed, this would work for me.  Personally I would prefer having an
output directory or --inplace as it makes easier for me (in my use case
I throw away the original files once  they have been scour'ed).

A third alternative is to have scour handle glob'ing itself.  That is,
you pass a literal *.svg to scour (e.g. \*.svg) and have scour expand it
internally.  Similar to many of git's commands.  That said, I fear
people will get this wrong more often than letting the shell expand the
glob.

> I'm also thinking about what should happen if a user calls scour like `scour *.svg`.
> 

If there is exactly two files, the user will trash one of the files
(with or without my proposed change).  As mentioned, I believe my
proposal handles that particular case (as the "last file" is unlikely to
be a directory).
  However, regardless of what route we choose, I think we should
definitely protect the user from accidentally trashing half of their svg
files if there are more than two.

IOW, if the user typed that (and there were more than two SVG files),
then scour ought to say "Did you want to use --in-place?" or "Missing
<option> to do bulk processing".

(The reason why I do not cover two files is that it is impossible for
scour to tell "scour in.svg out.svg" from "scour *.svg" apart if "*.svg"
expands to "in.svg out.svg").

nthykier:
Eduard Braun:
> #### Technical detail
> Whatever we decide regarding how to specifiy input/output files it should not influence the actual implementation (i.e. `references_relative_to` needs to go).
> 

Ok.  If I understand correctly, this implies that either:

  1. Embedding of raster images in bulk mode cannot be supported
     for now, OR
  2. There is a hard dependency on separating "command line options"
     from "scour instance options" (the part you mentioned below).

Is that correctly understood and would you be fine with option 1 in the
short term?

> What we should do is figuring out `options.infilename` and `options.outfilename` early (probably as early as `run()` if we want per-file statistics) and keep the rest of the current processing logic intact. To better understand what kind of separation I have in mind consider being able to process multiple files in parallel as a long-term goal.
> 

I can certainly appreciate parallel processing.  Not sure how much it
will gain us before CPython gets rid of its GIL.  However, it would
require that the code relied a lot less on global state, which has a lot
of other advantages.  Plus, I still hope we will one that get a GIL-less
python (without a major performance loss in single-threaded performance).

> What could be considered inconsistent in the current implementation is that the `options` object is both the location for where to store command line arguments as well as the "internal" options structure which should eventually be used for each "instance" of Scour - which is not ideal but probably an unnecessary hassle to split up right now, as long as it's clear that at some point each "instance" of Scour would get an own `options` object, operate on it, and eventually return a (yet to be created) `statistics` object with the result of the operation.
> 

If it is a hard dependency to get this resolved before scour can do bulk
processing, then I think we should look into this sooner rather than later.

oberstet:
Here are some thoughts from my side: processing many files with scour is "slow", because of 2 reasons:

a) Python startup time and
b) serial execution

a) is actually quite good with CPython (we should probably note in the docs that running Scour on PyPy works, but doesn't make sense), and it is not something we can do a lot about - other than processing multiple files in one run.

however, I am not sure if it is worth, because b) can be solved (and often is) using `make -jN`.

IMO, we should first look at how Scour is actually used in context. Eg if I am using it as part of a build process, then the build tool I use should be able to parallelize invocations of Scour. Eg if used from within Inkscape, as an exporter, @Ede123 will know best, but with such interactive use, I wouldn't expect a Scour internal parallelization to be important.

So actually, I am not convinced of the _why_, and for what goal / in what context ..

oberstet:
one more though: if this feature would be realized (batch processing, without restarting Python), then this could make PyPy interesting for Scour.

Eg PyPy is a tracing JIT that brings _massive_ performance increases, but only after some time of warmup (when hot code has been JITted).

In practice, it'll need to run for 20-30s at least Scour'ing files until it will speed up .. dramatically then.

It is useless when running for <10s .. the JIT won't kick in, and it will be slower than on CPy overall (total runtime).

nthykier:
It sounds like there is not support for this change as it is and accordingly, I am closing this pull request.

PR #181, opened 2018-03-18
Rename testscour.py to test_scour.py to make py.test work out of the box

nthykier:
This rename makes py.test/py.test-3 find the test suite out of the
box.  Example command lines:

       # Running the test suite
       $ py.test-3
       # Running the test suite with coverage enabled (and branch
       # coverage).
       $ py.test-3 --cov=scour --cov-report=html --cov-branch


Ede123:
Is there somewhere an explanation what pytest improves upon standard Python functionality?

I always dislike packages that seem to do what is already possible (maybe just in a different often only slightly more convenient way).

Long term I'd prefer to split the huge testscour.py into smaller modules and put them into a test/ subdirectory, but I did not get around to that yet...

Sorry for slow response times, I'm pretty busy these days, so it might take some time before I'm able to look into the remaining PRs (I'll look into them eventually, though, I promise).

nthykier:
Eduard Braun:
> Is there somewhere an explanation what pytest improves upon standard Python functionality?
>> I always dislike packages that seem to do what is already possible
(maybe just in a different often only slightly more convenient way).
> 

Feel free to reject the patch if you do not like the change. :)

I proposed it because it could be implemented side-by-side with scour's
own test runner while helping myself ("py.test-3" is my
"finger-memorized" test runner when I experiment with code/do iterative
changes locally).

> Long term I'd prefer to split the huge testscour.py into smaller modules and put them into a test/ subdirectory, but I did not get around to that yet...
> 

Do you have any public notes about what you were planning/expecting from
the refactoring?  Like how you wanted to group the tests and the
fixtures/input files.

> Sorry for slow response times, I'm pretty busy these days, so it might take some time before I'm able to look into the remaining PRs (I'll look into them eventually, though, I promise).
> 

Ok, take your time. :)


Thanks,
~Niels

Ede123:
> Do you have any public notes about what you were planning/expecting from the refactoring? Like how you wanted to group the tests and the fixtures/input files.

No, everything that exists is basically in my head and not developed any more than:
* I want a "tests" directory (not sure in which folder - might belong into /scour)
* This directory contains files like "test_xyz.py" which is basically the curent "testscour.py" split by topic, i.e. "xyz" will be things like "cli", "css", "ids", "path", ...
* Each test file can be run on it's own, by default however all tests are executed automatically

If you want to look into this feel free to open a dedicated issue where I'd be happy to discuss any ideas with you.

oberstet:
I am +1 on merging this (when it no longer conflicts;) - moving to pytest:

* **pytest** is widely used, best practice in py community (it has less boilerplate than unittest, it has test fixtures)
* besides pytest, I'd consider these pretty standard tools: **tox**, **coverage** and **mock**

nthykier:
Regarding the merge conflict.  I will gladly rebase it if @Ede123 agrees to the change.  However, given that every pull request that touches the test suite will cause a new conflict, I am hesitant to update it now.

Ede123:
@nthykier If you still want to merge this, could you please rebase and rename "testcss.py" -> "test_css.py" as mentioned above?

nthykier:
Rebased. It seems to work with the rename of `testcss.py` as well. :)

Ede123:
Thanks! :-)

Issue #184, opened 2018-04-08
slow (does not finish within one hour)


JoKalliauer:
Prozessing files over 10MB are very slow, and does not finish within several hours.

first example (contains 483.840 lines, 28MB)
https://upload.wikimedia.org/wikipedia/commons/f/ff/1_42_polytope_7-cube.svg
[Rendered by Wikipedia](https://commons.wikimedia.org/wiki/File:1_42_polytope_7-cube.svg)

Second example (70MB)
https://upload.wikimedia.org/wikipedia/commons/archive/9/9d/20171030170157%21E8out96g16.svg
[Rendering by Wikipedia](https://commons.wikimedia.org/wiki/File:E8out96g16.svg) (after optimisation/pathmerging with [svgcleaner](https://github.com/RazrFalcon/svgcleaner) ans [svgo](https://github.com/svg/svgo))

I ran the first file in Feb18 and the second in Oct17 over night wihtout success. (So they were tested with old versions of scour, but also in the actuall version it is more than 1000times slower than [svgcleaner](https://github.com/RazrFalcon/svgcleaner).)

If you want to optimise bulk-processing of big files I recommend to download the svg-files from:
https://commons.wikimedia.org/wiki/Category:Large_SVG_files

To not download them separately you can use Imker that downloads al files in a chosen category, (f.e (Large_SVG_files)[https://commons.wikimedia.org/wiki/Category:Large_SVG_files]) for you
https://commons.wikimedia.org/wiki/Commons:Imker_(batch_download)

nthykier:
For the first one (1_42_polytope_7-cube.svg), I see a runtime of about 8 minutes with python3 on master@6ea126d290e29c990244d9df17f00f84b7c07108.  While far from optimal, it seem like it is now doable to process and we are looking at some constant time improvements.

The 1_42_polytope_7-cube.svg consists almost exclusively on "<line ..>" rules.  Of the 491713 lines in the image, 483840 starts with `<line` (remaining items appear to be a few thousand `<circle>`-entries and a few "<g>` wrapping the other tags).

I have not have time to look at the other image in details yet.

nthykier:
The number of each type of tag (by looking only at the start of each line) in the the 70MB image:

```
     12 <ellipse
     54 <text
     91 <clipPath
     92 <g
     92 </g>
    105 <polyline
 142351 <path
 355157 <polygon
```

(I omitted the items that only appeared once in the above)

[...]

PR #185, opened 2018-04-10
Gh 184 perf faster remove default attr value

nthykier:
This reduces runtime by ~15-25% on the 1_42_polytope_7-cube.svg case provided in #184 (tested with python3).  Note that the runtime is still 5-6 minutes after this, so I would not consider #184 closed with just this commit.

### DIFF
--- a/scour/scour.py
+++ b/scour/scour.py
@@ -1876,6 +1876,20 @@ default_attributes = [
     DefaultAttribute('yChannelSelector', 'A', elements='feDisplacementMap')
 ]
 
+default_attributes_restricted_by_tag = defaultdict(list)
+default_attributes_unrestricted = []
+
+for attr in default_attributes:
+    if attr.elements is None:
+        # Applies to all tags
+        default_attributes_unrestricted.append(attr)
+        continue
+    if type(attr.elements) is str:
+        default_attributes_restricted_by_tag[attr.elements].append(attr)
+    else:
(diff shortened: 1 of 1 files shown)


Ede123:
Thanks for the improvement, merged now!

Turns out that after specifying valid elements for one residual attribute in 8ddb7d8913f862006cdc99d82dcb97e4f209f418 the list of default attributes valid for all elements is actually empty!

I'm thinking about reasons to keep the logic around, but I think we could actually drop the list completely and use the element-specific dict exclusively.   
The only reason I can think of right now to keep it around would be for users who wish to add default attributes of their own as we're unlikely to ever remove attributes from elements when we're not sure if they actually apply...

nthykier:
Eduard Braun:
> Thanks for the improvement, merged now!
> 

Thanks for the merge. :)

> Turns out that after specifying valid elements for one residual attribute in 8ddb7d8913f862006cdc99d82dcb97e4f209f418 the list of default attributes valid for all elements is actually empty!
> 

Nice.  :)

> I'm thinking about reasons to keep the logic around, but I think we could actually drop the list completely and use the element-specific dict exclusively.   
> The only reason I can think of right now to keep it around would be for users who wish to add default attributes of their own as we're unlikely to ever remove attributes from elements when we're not sure if they actually apply...
> 

Ack.  I will leave that call to you. :)

Thanks,
~Niels

Issue #186, opened 2018-04-12
running scour multiple times, leads every second time to the same size


JoKalliauer:
I run scour multiple times since it reaches a static size, but I already had several files where scour switches between two (or three) different sizes.

I sometimes run scour with all options enabled.
> scour -i input.svg -o output.svg --enable-comment-stripping --remove-titles --remove-descriptions --disable-embed-rasters --strip-xml-space  --create-groups --remove-metadata --remove-descriptive-elements --renderer-workaround  --enable-viewboxing --enable-id-stripping --shorten-ids --set-precision=5 --set-c-precision=3

373750 bytes:
[First.svg.txt](https://github.com/scour-project/scour/files/1905018/First.svg.txt)

373560 bytes:
[Second.svg.txt](https://github.com/scour-project/scour/files/1905019/Second.svg.txt)

373750 bytes:
[Third.svg.txt](https://github.com/scour-project/scour/files/1905020/Third.svg.txt)

Copyright: Public Domain by grnsgrf
scource: https://commons.wikimedia.org/wiki/File:Welcome_to_Fabulous_Las_Vegas_Sign.svg

nthykier:
Based on a wdiff, the concrete case appears to be caused by --shorten-ids.  The IDs are not stable and are apparently not always optimal.

nthykier:
@JoKalliauer - which version of scour were you using for this test? 0.36?

nthykier:
Hmm, v0.35 does not recognise --set-c-precision=3 and I cannot reproduce it with a git checkout of v0.36 nor master@6ea126d290e29c990244d9df17f00f84b7c07108.  For me, scour on the first fine consistently generate the same output (bit-for-bit identical) even if I change version (v0.36 vs. master) or python (2.7 vs. 3.6.5rc1)

Ede123:
The problem is that IDs get are generated in a way that gives elements with the most references the shortest IDs. When running the second time on an already processed file the shortest IDs are therefore already "taken" and replaced with potentially longer ones.

What we need to do is assign "temporary" IDs to all elements and/or exchange IDs when necessary, but it needs some investigation on how to do this efficiently.

nthykier:
Ack, I am playing with the code right now to see if I can solve it.

nthykier:
Btw, with this particular input file, I see a factor ~7 different in runtime depending on python version (py2: ~16s, py3: ~2.5s).

[...]

PR #187, opened 2018-04-13
Enable shortenIDs to recycle existing IDs

nthykier:
This patch enables shortenIDs to remap IDs currently in use.  This is
very helpful to ensure that scour does not change been "optimal" and
"suboptimal" choices for IDs as observed in GH#186.

Closes: #186

### DIFF
--- a/scour/scour.py
+++ b/scour/scour.py
@@ -700,6 +700,10 @@ def shortenIDs(doc, prefix, options):
     num = 0
 
     identifiedElements = findElementsWithId(doc.documentElement)
+    # This map contains maps the (original) ID to the nodes referencing it.
+    # At the end of this function, it will no longer be valid and while we
+    # could keep it up to date, it will complicate the code for no gain
+    # (as we do not reuse the data structure beyond this function).
     referencedIDs = findReferencedElements(doc.documentElement)
 
     # Make idList (list of idnames) sorted by reference count
@@ -734,7 +738,7 @@ def shortenIDs(doc, prefix, options):
         # (e.g. remapping "c" to "a" is not going to win us anything)
         if len(curId) != len(rid):
             # Then go rename it.
-            num += renameID(doc, rid, curId, identifiedElements, referencedIDs)
+            num += renameID(doc, rid, curId, identifiedElements, referencedIDs.get(rid))
         elif curId < rid:
             # If we skip reassigning an ID because it has the same length
             # (E.g. skipping "c" -> "a"), then we have to mark the "future"
@@ -765,14 +769,12 @@ def intToID(idnum, prefix):
     return prefix + rid
 
 
-def renameID(doc, idFrom, idTo, identifiedElements, referencedIDs):
+def renameID(doc, idFrom, idTo, identifiedElements, referringNodes):
     """
     Changes the ID name from idFrom to idTo, on the declaring element
-    as well as all references in the document doc.
+    as well as all nodes in referringNodes.
 
-    Updates identifiedElements, but not referencedIDs.
-    Does not handle the case where idTo is already the ID name
-    of another element in doc.
+    Updates identifiedElements.
 
     Returns the number of bytes saved by this replacement.
     """
@@ -786,7 +788,6 @@ def renameID(doc, idFrom, idTo, identifiedElements, referencedIDs):
     num += len(idFrom) - len(idTo)
 
     # Update references to renamed node
-    referringNodes = referencedIDs.get(idFrom)
     if referringNodes is not None:
 
         # Look for the idFrom ID name in each of the referencing elements,
@@ -836,11 +837,6 @@ def renameID(doc, idFrom, idTo, identifiedElements, referencedIDs):
                     newValue = newValue.replace('url("#' + idFrom + '")', 'url(#' + idTo + ')')
                     node.setAttribute(attr, newValue)
                     num += len(oldValue) - len(newValue)
-
-        # We deliberately leave referencedIDs alone to enable us to bulk update
-        # IDs where two nodes swap IDs.  (GH#186)
-        # del referencedIDs[idFrom]
-        # referencedIDs[idTo] = referringNodes
 
     return num
 
--- a/scour/scour.py
+++ b/scour/scour.py
@@ -690,19 +690,16 @@ def removeUnreferencedElements(doc, keepDefs):
     return num
 
 
-def shortenIDs(doc, prefix, unprotectedElements=None):
+def shortenIDs(doc, prefix, options):
     """
     Shortens ID names used in the document. ID names referenced the most often are assigned the
     shortest ID names.
-    If the list unprotectedElements is provided, only IDs from this list will be shortened.
 
     Returns the number of bytes saved by shortening ID names in the document.
     """
     num = 0
 
     identifiedElements = findElementsWithId(doc.documentElement)
-    if unprotectedElements is None:
-        unprotectedElements = identifiedElements
     referencedIDs = findReferencedElements(doc.documentElement)
 
     # Make idList (list of idnames) sorted by reference count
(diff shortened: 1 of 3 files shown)


oberstet:
@nthykier oh, cool: I just noticed that you have a number of PRs queuing .. thanks for contributing! @Ede123 how is your current time budget? =) if none, I see what I can do @nthykier to review it

nthykier:
Tobias Oberstein:
> @nthykier oh, cool: I just noticed that you have a number of PRs queuing .. thanks for contributing!

No problem.  I am glad that some of these changes are useful :)

Thanks,
~Niels

scour/scour.py
```diff
@@ -822,34 +823,44 @@ def renameID(doc, idFrom, idTo, identifiedElements, referencedIDs):
...
-        referencedIDs[idTo] = referringNodes
+        # We deliberately leave referencedIDs alone to enable us to bulk update
+        # IDs where two nodes swap IDs.  (GH#186)
+        # del referencedIDs[idFrom]
+        # referencedIDs[idTo] = referringNodes
 
     return num
 
 
+def _protectedIDs(seenIDs, options):
```
Ede123: You should name this `protected_ids` for consistency with the other function (which I feel is more important than to mark this internal, for which the long-term goal is something along the lines of #121).

Also note we try to use PEP8 naming for new code where reasonable (don't force it though - only if it makes sense and doesn't introduce any inconsistencies). Naming similar things similarly is always more important than following any guidelines.

nthykier:
Eduard Braun:
> Ede123 requested changes on this pull request.
> 
> General idea and most of the code looks fine, however it causes a serious regression:

Hi,

Thanks for the review. :)

I have pushed fixes as separate commits on top of the branch.  Let me
know if (and in that case, when) you want the branch rebased/squashed.

> Since we can now re-use IDs already present in the document we need to make sure to properly rename them in the process (I'm not sure what the best possibility for this is, especially in a way that does not potentially regress performance - i.e. consider a case where you add one heavily referenced object to an already scoured document which would require Scour to rename all IDs - so this part needs some investigation).
> 

Added in d6406a34706e7ae1422595a5eda6d8bc7b50b1a8.

To be honest, I think the major win here is making scour produce
"stable" output once it has picked shortened IDs.  The performance
optimization is only applicable in a "best case" scenario where optimal
IDs have already been handed out.

Technically, I have solved it by ensuring that scour does not change the
ID of a node unless the ID changes size.

> Testcase:
> ```
> [...]
> ```
> 

Thanks for that.  I have added a unit test using that to avoid regressions.

>> @@ -822,34 +823,44 @@ def renameID(doc, idFrom, idTo, identifiedElements, referencedIDs):
>                      node.setAttribute(attr, newValue)
>                      num += len(oldValue) - len(newValue)
>  
> -        del referencedIDs[idFrom]
> -        referencedIDs[idTo] = referringNodes
> +        # We deliberately leave referencedIDs alone to enable us to bulk update
> +        # IDs where two nodes swap IDs.  (GH#186)
> +        # del referencedIDs[idFrom]
> +        # referencedIDs[idTo] = referringNodes
> 
> I don't understand this comment, can you clarify?
> 

I have deleted it and have replaced it with
91503c6d7e0882ebfff7bb52def7938a63e9864e

>>  
>      return num
>  
>  
> +def _protectedIDs(seenIDs, options):
> 
> You should name this `protected_ids` for consistency with the other function (which I feel is more important than to mark this internal, for which the long-term goal is something along the lines of #121).
> 

Ok, done in 00cf42b554e11e09348218b9e66d72de178694ae

> Also note we try to use PEP8 naming for new code where reasonable (don't force it though - only if it makes sense and doesn't introduce any inconsistencies). Naming similar things similarly is always more important than following any guidelines.
> 

This seems like a tricky balance that I am sure I will get wrong again
eventually. :)

Thanks,
~Niels

Ede123:
>  Technically, I have solved it by ensuring that scour does not change the ID of a node unless the ID changes size.

Sounds reasonable. It does not really matter wether we assign "a" or "z" as an ID from a technical point of view.  
The only downside I can think of is that scouring the original file (with a small modification) vs. scouring a previously scoured file (with the same small modification) would result in different IDs, but I assume that's hardly relevant in practice.

> Thanks for that. I have added a unit test using that to avoid regressions.

I'm afraid I did not describe the regression well enough (and my testcase wasn't designed well enough either). The problem occurs when actually re-writing is in order, see this updated minimal testcase, which still reveals the behavior I wanted to show before:

```SVG
<?xml version="1.0" encoding="UTF-8"?>
<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
 <defs>
  <rect id="a" width="80" height="50" fill="red"/>
  <rect id="long_id" width="80" height="50" fill="blue"/>
 </defs>
 <use xlink:href="#a"/>
 <use xlink:href="#long_id"/>
 <use xlink:href="#long_id"/>
</svg>
```

The "long_id" gets rewritten to "a"!

Yet another testcase which might be close to "real world" files which were scoured and later modified: [testcase.svg.txt](https://github.com/scour-project/scour/files/1913258/testcase.svg.txt)  
In this case the new algorithm also rewrites "long_id" "a" and even manages to update the previous link to "a" with "aa", but fails to update the actual IDs, so it effectively exchanges the objects.

> This seems like a tricky balance that I am sure I will get wrong again eventually. :) 

Well, in general PEP8 names are the way we want to go, so if in doubt, use them. I'm only hesitant to rewrite most of the code just to follow PEP8 names (changes like these make it unnecessarily hard to browse the git history using "git blame" while not really improving anything).

nthykier:
Eduard Braun:
>>  Technically, I have solved it by ensuring that scour does not change the ID of a node unless the ID changes size.
> 
> Sounds reasonable. It does not really matter wether we assign "a" or "z" as an ID from a technical point of view.  

Indeed.  That is what I thought.

By the way, I noticed that scour's ID allocator only uses a-z, but
according the the SVG standard (via the XML standard) allows for a lot
more characters in the ID field.

https://www.w3.org/TR/2008/REC-xml-20081126/#NT-Name says:

> [4]   	NameStartChar	   ::=   	":" | [A-Z] | "_" | [a-z] | [#xC0-#xD6] | [#xD8-#xF6] | [#xF8-#x2FF] | [#x370-#x37D] | [#x37F-#x1FFF] | [#x200C-#x200D] | [#x2070-#x218F] | [#x2C00-#x2FEF] | [#x3001-#xD7FF] | [#xF900-#xFDCF] | [#xFDF0-#xFFFD] | [#x10000-#xEFFFF]
> [4a]   	NameChar	   ::=   	NameStartChar | "-" | "." | [0-9] | #xB7 | [#x0300-#x036F] | [#x203F-#x2040]
> [5]   	Name		   ::=   	NameStartChar (NameChar)*

(from https://www.w3.org/TR/2008/REC-xml-20081126/#id, which is found
via https://www.w3.org/TR/SVG/struct.html#Core.attrib)

In other words, we can at least use "_", uppercase letters.
Furthermore, if we handle the first character differently from the rest,
then we can use 0-9, "-" and "." as well (without having to look at all
the UTF-8 characters also permitted).


> The only downside I can think of is that scouring the original file (with a small modification) vs. scouring a previously scoured file (with the same small modification) would result in different IDs, but I assume that's hardly relevant in practice.
> 

In the general case, we cannot optimally shorten IDs and still produce
identical output if people alter the reference counts of the image
between runs of scour.

Obviously, we can still avoid "pointless" reassignments or the
"avalanche-effect" when we assign an existing ID a shorter name.

>> Thanks for that. I have added a unit test using that to avoid regressions.
> 
> I'm afraid I did not describe the regression well enough (and my testcase wasn't designed well enough either). The problem occurs when actually re-writing is in order, see this updated minimal testcase, which still reveals the behavior I wanted to show before:
> 
> ```SVG
> <?xml version="1.0" encoding="UTF-8"?>
> <svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
>  <defs>
>   <rect id="a" width="80" height="50" fill="red"/>
>   <rect id="long_id" width="80" height="50" fill="blue"/>
>  </defs>
>  <use xlink:href="#a"/>
>  <use xlink:href="#long_id"/>
>  <use xlink:href="#long_id"/>
> </svg>
> ```
> 
> The "long_id" gets rewritten to "a"!
> 

Thanks for the extended test case. :)

> Yet another testcase which might be close to "real world" files which were scoured and later modified: [testcase.svg.txt](https://github.com/scour-project/scour/files/1913258/testcase.svg.txt)  
> In this case the new algorithm also rewrites "long_id" "a" and even manages to update the previous link to "a" with "aa", but fails to update the actual IDs, so it effectively exchanges the objects.
> 

Ok, will have a look at fixing this issue.

Thanks,
~Niels

nthykier:
@Ede123 I have written 039022e, which I believe addresses your concerns.

nthykier:
Hey @Ede123

Let me know if there is anything I can do to improve the changes in this pull request.

JoKalliauer:
Thanks to @nthykier : I can confirm it is fixed.

Issue #190, opened 2018-04-17
IndexError: list index out of range on SVG 1.1 reference file


nthykier:
scour on the current master branch fails to process several of the SVG 1.1 reference SVGs. Below is one of these failures:

```
Processing painting-marker-01-f.svg
Traceback (most recent call last):
  File "scour.py", line 3, in <module>
    run()
  File "$GIT_DIR/scour/scour.py", line 3928, in run
    start(options, input, output)
  File "$GIT_DIR/scour/scour.py", line 3896, in start
    out_string = scourString(in_string, options).encode("UTF-8")
  File "$GIT_DIR/scour/scour.py", line 3570, in scourString
    _num_bytes_saved_in_transforms = optimizeTransforms(doc.documentElement, options)
  File "$GIT_DIR/scour/scour.py", line 3020, in optimizeTransforms
    num += optimizeTransforms(child, options)
  File "$GIT_DIR/scour/scour.py", line 3020, in optimizeTransforms
    num += optimizeTransforms(child, options)
  File "$GIT_DIR/scour/scour.py", line 3020, in optimizeTransforms
    num += optimizeTransforms(child, options)
  File "$GIT_DIR/scour/scour.py", line 3007, in optimizeTransforms
    optimizeTransform(transform)
  File "$GIT_DIR/scour/scour.py", line 2969, in optimizeTransform
    if prevArgs[0] == prevArgs[1] == 1:
IndexError: list index out of range
```

Options used:
```sh
scour --enable-id-stripping --enable-comment-stripping --shorten-ids --indent=none  --no-line-breaks --strip-xml-prolog --remove-descriptive-elements --set-precision=8 --set-c-precision=8 --create-groups --remove-titles --remove-descriptions --remove-metadata
```

Test data is from: https://www.w3.org/Graphics/SVG/Test/20110816/

PR #191, opened 2018-04-18
Avoid crashing on "scale(1)" (short for "scale(1, 1)")

nthykier:
The scale function on the transform attribute has a short form, where
only the first argument is used.  But optimizeTransform would always
assume that there were two when checking for the identity scale.

Closes: #190

### DIFF
--- a/scour/scour.py
+++ b/scour/scour.py
@@ -2966,7 +2966,9 @@ def optimizeTransform(transform):
                 # y1 * uniformscalefactor2
                 prevArgs[1] *= currArgs[0]
             del transform[i]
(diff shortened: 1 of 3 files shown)


nthykier:
The "scale(1) == scale(1 1)" assertion is from: https://www.w3.org/TR/SVG11/single-page.html#coords-TransformAttribute

Quote:
```
scale(<sx> [<sy>]), which specifies a scale operation by sx and sy. If <sy> is not provided, it is assumed to be equal to <sx>.
```

oberstet:
this looks good and it has test coverage => merged.

nthykier:
Thanks for the merge. :)

PR #199, opened 2018-07-01
Several improvements for handling whitespace including newlines

Ede123:


scour/scour.py
```diff
@@ -3341,19 +3341,30 @@ def serializeXML(element, options, indent_depth=0, preserveWhitespace=False):
...
-                if preserveWhitespace:
-                    outParts.append(makeWellFormed(child.nodeValue))
-                else:
-                    outParts.append(makeWellFormed(child.nodeValue.strip()))
+                text_content = child.nodeValue
+                if not preserveWhitespace:
+                    # strip / consolidate whitespace according to spec, see
+                    #    https://www.w3.org/TR/SVG/text.html#WhiteSpace
+                    # As a workaround for inconsistent handling of renderers keep newlines if they were in the original
+                    if element.nodeName in ['text', 'tspan', 'tref', 'textPath', 'altGlyph']:
```
nthykier: Looks like the if-body could be replaced by:

```python
    text_content = ws_norm_pattern.sub(' ', text_content).strip(' ')
```

With ws_norm_pattern being defined as:

```python
ws_norm_pattern = re.compile('[ \t]+')
```

outside the function (so we do not recompile the pattern all the time)

It also has the advantage that is avoids an `O(n^2)` update of the string in text_content for strings with a lot of whitespace in the `while '  ' in text_content: ...`-loop.

Ede123: If regular expressions are fast enough I'm fine with that (I guess I'll do some measurements unless you already know the result).

Just to warn you: I don't like the second commit yet, so I'm still going to rewrite some parts.

Ede123: @nthykier OK, changes were minimal (basically I decided to discard my previous idea and strip newlines unconditionally (as suggested in the spec but not applied by Firefox unfortunately).

Regarding usage of regular expressions: I did some tests and it turns out `string.replace()` is usually faster than a re. Especially for short strings (and we have to consider white-space only text nodes here, too!) it clearly wins. Even for longs strings the regular expressions does not give an advantage. Only for examples that shouldn't be common in practice (long strings with many consecutive spaces) the re seems to give a small advantage. Feel free to cross-check my findings if you want (maybe even on some real world data)!

Please note we can *not* call `.strip(' ')` unconditionally, as for example `<text>text <tspan>tspan</tspan> text</text>` should result in "text text text" as the whole "character data" of the text node is to be processed in one step and not as individual childs (that's what I gathered from an IRC discussion I had with Tavmjong Bah on IRC yesterday).

nthykier: Ok. :)

PR #208, opened 2018-08-18
Merge sibling <g> nodes with identical attributes

nthykier:
In some cases, gnuplot generates a very suboptimal SVG content of the
following pattern:

        <g color="black" fill="none" stroke="currentColor">
        <path d="m82.5 323.3v-4.1" stroke="#000"/>
        </g>
        <g color="black" fill="none" stroke="currentColor">
        <path d="m116.4 323.3v-4.1" stroke="#000"/>
        </g>
        ... repeated 10+ more times here ...
        <g color="black" fill="none" stroke="currentColor">
        <path d="m65.4 72.8v250.5h420v-250.5h-420z" stroke="#000"/>
        </g>

A more optimal pattern would be:

        <g color="black" fill="none" stroke="#000">
        <path d="m82.5 323.3v-4.1"/>
        <path d="m116.4 323.3v-4.1"/>
        ... 10+ more paths here ...
        <path d="m65.4 72.8v250.5h420v-250.5h-420z"/>
        </g>

This patch enables that optimization by handling the merging of two
sibling <g> entries that have identical attributes.  In the above
example that does not solve the rewrite from "currentColor" to "#000"
for the stroke attribute.  However, the existing code already handles
that automatically after the <g> elements have been merged.

This change provides comparable results to --create-groups as shown by
the following diagram while being a distinct optimization:

       +----------------------------+-------+--------+
       |           Test             | Size  |  in %  |
       +----------------------------+-------+--------+
       | baseline                   | 17961 |  100%  |
       | baseline + --create-groups | 17418 |  97.0% |
       | patched                    | 16939 |  94.3% |
       | patched + --create-groups  | 16855 |  93.8% |
       +----------------------------+-------+--------+

The image used in the size table above was generated based on the
instructions from https://bugs.debian.org/858039#10 with gnuplot 5.2
patchlevel 2.  Beyond the test-based "--create-groups", the following
scour command-line parameters were used:
      --enable-id-stripping --enable-comment-stripping \
      --shorten-ids --indent=none

Note that the baseline was scour'ed repeatedly to stablize the image
size.

### DIFF
--- a/scour/scour.py
+++ b/scour/scour.py
@@ -1134,6 +1134,75 @@ def moveCommonAttributesToParentGroup(elem, referencedElements):
     return num
 
 
+def mergeSiblingGroupsWithCommonAttributes(elem):
+    """
+    Merge two or more sibling <g> elements with the identical attributes.
+
+    This function acts recursively on the given element.
+    """
+
+    num = 0
+    i = elem.childNodes.length - 1
+    while i >= 0:
+        currentNode = elem.childNodes.item(i)
+        if currentNode.nodeType != Node.ELEMENT_NODE or currentNode.nodeName != 'g' or \
+           currentNode.namespaceURI != NS['SVG']:
+            i -= 1
+            continue
+        attributes = {a.nodeName: a.nodeValue for a in currentNode.attributes.values()}
+        runStart, runEnd = i, i
+        runElements = 1
+        while runStart > 0:
+            nextNode = elem.childNodes.item(runStart - 1)
+            if nextNode.nodeType == Node.ELEMENT_NODE:
+                if nextNode.nodeName != 'g' or nextNode.namespaceURI != NS['SVG']:
+                    break
+                nextAttributes = {a.nodeName: a.nodeValue for a in nextNode.attributes.values()}
+                hasNoMergeTags = (True for n in nextNode.childNodes
+                                  if n.nodeType == Node.ELEMENT_NODE
+                                  and n.nodeName in ('title', 'desc')
+                                  and n.namespaceURI == NS['SVG'])
+                if attributes != nextAttributes or any(hasNoMergeTags):
+                    break
+                else:
+                    runElements += 1
+                    runStart -= 1
+            else:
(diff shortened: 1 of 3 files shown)


Ede123:
Sorry for the *very* late reply.

Looks good in principle but raises some questions:
* Could you add a simple test to verify the functionality is working?
* Should `mergeSiblingGroupsWithCommonAttributes` be called after `createGroupsForCommonAttributes` or is the current order more suitable?
* Should this option be affected by `options.group_create` or `options.group_collapse` respectively? It *does* collapse groups in a sense, and it *does* create new groups in a sense.

nthykier:
Hi,

I have rebased the branch against current master, including a test case and an two bug fixes. One of the bug fix seems to have improved its efficiency (the commit message has not been updated to reflect that though).

As for the questions:

 * I put merge before create because it merge can eliminate slightly better than create can in the common cases I was going for.  If you are going for optimal results, the two should probably be merged (kind of like how some the compiler optimizations for constant folding and branch elimination are better when combined than as two individual optimizations).  However, I do not think that is worth the hassle at the moment (specially because merging has no risk of creating larger files but create groups can in some cases).

 * From my point of view, merging is neither creating or collapsing nodes.  Though of the two, I think `--disable-group-collapsing` is better/closest.  Furthermore, my aim was to create an optimization that produced good results with no risk of expanding the file size (`--create-groups` can do that in some cases). I would prefer for it to be independent of ` --create-groups` to ensure it is enabled more often.

Ede123:
Thanks, sounds good!

Let's make it take `--disable-group-collapsing` into account, because it *does* destroy groups that existed before. Personally I only disable group collapsing in files where I a) know I already created optimal groups or b) use the groups for logic grouping of elements, so I don't want scour to mess with them.

nthykier:
Ack, I have rebased with this code path being conditional on `--disable-group-collapsing` (along with a test case to assert that as well).

Ede123:
Great, thanks!

Issue #215, opened 2018-10-22
group-collapsing


JoKalliauer:
Running scour for [InputOutput.svg.txt](https://github.com/scour-project/scour/files/2503085/InputOutput.svg.txt) does not collapse groups.

scour -i [InputOutput.svg.txt](https://github.com/scour-project/scour/files/2503085/InputOutput.svg.txt)  -o [InputOutput.svg.txt](https://github.com/scour-project/scour/files/2503085/InputOutput.svg.txt) 

If I understand [`--disable-group-collapsing`](https://github.com/scour-project/scour/wiki/Documentation-(first-draft)#--disable-group-collapsing) correctly then it should collapse groups by default.

# Input=Output
[InputOutput.svg.txt](https://github.com/scour-project/scour/files/2503085/InputOutput.svg.txt) 
```svg
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE svg  PUBLIC '-//W3C//DTD SVG 1.1//EN'  'http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd'>
<svg viewBox="0 0 1e3 1e3" xmlns="http://www.w3.org/2000/svg">
 <g font-family="Liberation Sans,Arial,Helvetica,sans-serif">
  <g text-anchor="middle">
   <g font-weight="400">
    <g font-size="64">
     <text x="99" y="99">Text</text>
    </g>
   </g>
  </g>
 </g>
</svg>
```

# desired output
[whish.svg.txt](https://github.com/scour-project/scour/files/2503090/whish.svg.txt)
```svg
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE svg  PUBLIC '-//W3C//DTD SVG 1.1//EN'  'http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd'>
<svg viewBox="0 0 1e3 1e3" xmlns="http://www.w3.org/2000/svg">
 <text x="99" y="99" font-family="Liberation Sans,Arial,Helvetica,sans-serif" font-size="64" font-weight="400" text-anchor="middle">Text</text>
</svg>
```

Or what does group collapsing mean? (maybe one example)

Perhelion:
One group would be also ok. For groups and text there are much more optimizations possible, e.g. including anti-groups and full transformation resolution.

Gibbz:
Yes id love to see support for removing all groups where possible

PR #244, opened 2020-05-17
WIP: Add optimization that prunes nested <g>-tags

nthykier:
An optimization that prunes nested <g>-tags when they contain exactly one <g> and nothing else (except whitespace nodes).  This looks a bit like `removeNestedGroups` except it only touches <g> tags without attributes (but can remove <g>-tags completely from a tree, whereas this optimization always leaves at least one <g> tag behind).

Closes: #215


nthykier:
(have seen the flake error, will look at it later)

nthykier:
I need to follow up on something before it can be merged.

PR #247, opened 2020-05-21
Perf optimized serialization

nthykier:
Some commits to reduce the overhead in outputting the SVG again (most of it is in `serializeXML` and below)

### DIFF
--- a/scour/scour.py
+++ b/scour/scour.py
@@ -74,6 +74,12 @@ VER = __version__
 COPYRIGHT = u'Copyright Jeff Schiller, Louis Simard, 2010'
 
 
+XML_ENTS_NO_QUOTES = {'<': '&lt;', '>': '&gt;', '&': '&amp;'}
+XML_ENTS_ESCAPE_APOS = XML_ENTS_NO_QUOTES.copy()
+XML_ENTS_ESCAPE_APOS["'"] = '&apos;'
+XML_ENTS_ESCAPE_QUOT = XML_ENTS_NO_QUOTES.copy()
+XML_ENTS_ESCAPE_QUOT['"'] = '&quot;'
+
 NS = {'SVG':      'http://www.w3.org/2000/svg',
       'XLINK':    'http://www.w3.org/1999/xlink',
       'SODIPODI': 'http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd',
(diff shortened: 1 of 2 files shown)


scour/scour.py
```diff
@@ -3433,6 +3433,42 @@ def choose_quote_character(value):
...
+    ]
+
+KNOWN_ATTRS_ORDER_BY_NAME = defaultdict(lambda: len(KNOWN_ATTRS),
+                                        {name: order for order, name in enumerate(KNOWN_ATTRS)})
+
+
+# use custom order for known attributes and alphabetical order for the rest
+def _attribute_sort_key_function(attribute):
+    name = attribute.name
+    order_value = KNOWN_ATTRS_ORDER_BY_NAME[attribute.name]
```
Ede123, line 3461: Use `name` instead of `attribute.name` here (or don't define `name` at all).

scour/scour.py
```diff
@@ -3433,6 +3433,42 @@ def choose_quote_character(value):
...
+    return order_value, name
+
+
+def attributes_ordered_for_output(elem):
+    if not elem.hasAttributes():
+        return []
+    attr = elem.attributes
```
Ede123, line 3468: Personal preference (i.e. no need to fix if doing so would annoy you 😉): Use full names, e.g.
* `element` instead of `elem`
* `attributes = element.attributes` instead of `attr = elem.attributes`

Especially when refactoring code, it's usually a good opportunity to implement such changes also to existing code.

nthykier:
> I assume you have benchmarked the result of each of these improvements somehow?
> 

Yes, though originally I benchmarked the result of the branches combined (i.e. #248 + #249 + this) and then split them into "related groups of optimizations" later to make reviewing and merging easier.

> Could you share these results with us somehow (I'm mostly curious and too lazy to do a comparison myself right now wink )

This branch gives 2.75% performance boost on the `secure-card.svgz` from the Oxygen icon theme.  In absolute values, this is on average 0.08s per run for that file (from `2.90s` to `2.81s`).

scour/scour.py
nthykier: Done

scour/scour.py
nthykier: Done.

scour/scour.py
```diff
@@ -3413,6 +3413,11 @@ def remapNamespacePrefix(node, oldprefix, newprefix):
 def make_well_formed(text, quote_dict=None):
     if quote_dict is None:
         quote_dict = XML_ENTS_NO_QUOTES
+    if not any(c in text for c in quote_dict):
+        # The quote-able characters are quite rare in SVG (they mostly only
+        # occur in text elements in practice).  Therefore it make sense to
+        # optimize for this common case
+        return text
     return ''.join(quote_dict[c] if c in quote_dict else c for c in text)
```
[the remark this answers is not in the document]
nthykier: It is a micro-optimization and I cannot think of anything else using pure python that would make the original statement go notably faster.  One of the wins here is to avoid allocate memory to clone a string when we can just return the original one.  Obviously, if we start to look at C or Cython, then our options expand (but so does the complexity of the optimization).

The only final option, I can think of (but did not test), was to use a cache mechanism to avoid triggering the case in the first place (a la `@lru_cache`). But if we go down that route, then `quote_dict` should probably no longer be an input parameter (to keep the caching simple and cheap).

It is relevant to know that we rarely ever need to use the escape/quote sequences in practise as most attributes have a well-defined set of values that can be inserted.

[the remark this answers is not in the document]
Ede123: Great, seems you already put a lot of thought into it, so I trust in your choice that this is the best way to go.

PR #248, opened 2020-05-21
Optimize remove duplicate gradients

nthykier:


scour/scour.py
```diff
@@ -1543,35 +1543,34 @@ def removeDuplicateGradients(doc):
...
-                    dup_id = dup.getAttribute('id')
+                for i in range(len(duplicates_ids)):
+                    dup_id = duplicates_ids[i]
                     if dup_id:
+                        # We do not bother updating the master field
+                        # as it is not used any more.
+                        master_id = duplicates_ids[i]
                         duplicates[i] = master
-                        master = dup
+                        duplicates_ids[i] = master_id
```
Ede123, line 1565: Wait a moment - do I see correctly that we assign like
```py
master_id = duplicates_ids[i]
duplicates_ids[i] = master_id
```
i.e. effectively `duplicates_ids[i] = duplicates_ids[i]`?

This can go, right? We only need the line `duplicates[i] = master`

nthykier: Thanks for finding that.

Partly, we also "need" to clear `duplicates_ids[i]` to avoid remapping the master id to itself.  The end result is the same, but we would avoid some unnecessary remappings.

nthykier: Done.

scour/scour.py
```diff
@@ -1609,16 +1609,36 @@ def removeDuplicateGradients(doc):
...
         prev_num = num
 
         linear_gradients = doc.getElementsByTagName('linearGradient')
         radial_gradients = doc.getElementsByTagName('radialGradient')
 
-        # get a collection of all elements that are referenced and their referencing elements
-        referenced_ids = findReferencedElements(doc.documentElement)
         for master_id, duplicates_ids, duplicates in detect_duplicate_gradients(linear_gradients, radial_gradients):
             dedup_gradient(master_id, duplicates_ids, duplicates, referenced_ids)
+            # If the gradients have an ID, we update referenced_ids to match the newly remapped IDs.
```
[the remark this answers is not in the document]
nthykier: Done.

scour/scour.py
```diff
@@ -1532,82 +1532,126 @@ def computeGradientBucketKey(grad):
...
+                _setStyle(elem, styles)
+
+        # now that all referencing elements have been re-mapped to the master
+        # it is safe to remove this gradient from the document
+        dup_grad.parentNode.removeChild(dup_grad)
+
+    # If the gradients have an ID, we update referenced_ids to match the newly remapped IDs.
+    # This enable us to avoid calling findReferencedElements once per loop, which is helpful as it is
+    # one of the slowest functions in scour.
+    if master_id:
```
Ede123, line 1621: One last question: You check for `master_id` here, but shouldn't it be guaranteed that we have a master ID here?

Or rather, if we *don't* have a master ID, doesn't that mean that none of the duplicates had an ID, therefore `detect_duplicate_gradients` shouldn't have returned the tuple in the first place, because there's nothing we can de-duplicate and in fact also nothing that can reference the gradient, so we probably could have dropped it right away?

nthykier: It might be the case that we can drop the gradient right away.  However, #203 tells us that gradients without IDs do appear when we attempt to deduplicate gradients in the current codebase.

My personal take would be to open a separate issue for why we are not pruning ID-less gradients earlier (with its own PR to fix that and remove this eventually redundant check).

Ede123: Ah, I thought that only happened for gradients that did not have an ID to start with. In this case we obviously need to investigate further.

Issue #23, opened 2015-12-06
Import history from old repository


Ede123:
I found a possibility ([`git-replace`](https://git-scm.com/docs/git-replace)) to import the full history from the Launchpad repository, **linking** it to the GitHub commit history **without rebasing** (all the recent commits remain unchanged including hashes!)

In fact my fork at https://github.com/Ede123/scour contains the full history right now, see https://github.com/Ede123/scour/commits/39fb9eee021e328261075481a2addbdca2c6cf19.

I'm afraid GitHub does not support replacements in its front-end, so it will not show the combined history. However if one clones the repository locally and executes `git fetch origin 'refs/replace/*:refs/replace/*'` the full history will be available and visible when doing e.g. `git log`

Since GitHub does not allow me to create a PR for this (since there are obviously no changes when comparing):
Shall I describe the necessary steps in detail so someone with write access to the repository can re-do the import?

oberstet:
To be honest, I have never done that before, but I have been bitten by Git tricks modifying history before. So I'm a bit sceptical.

@Ede123 Could you describe the steps that you did? How would be try that here without risking screwing up the whole repo?

@codedread Is this (having complete history here) something you would see as important? You started this code base, so I think you should have final decision on this.

Ede123:
First of all: My approach does not modify history **at all**! That's the beauty of `git-replace`. The current commits remain completely unchanged. The commits from the history are simply added in parallel (i.e. have no connection to the current commits).
Only by using `git-replace` a virtual link between the two separate commit histories is created. This is done by *virtually* replacing one single commit of the current commits with a modified version that includes the the last commit of the old history as a parent.
If one does not fetch this replacement (which is stored in `.git/refs/replace/` as a plain text file) with the command I posted above absolutely nothing will change. If one does fetch the replacement one will have the combined commit history.

For the whole procedure I first followed [this guide](https://astrofloyd.wordpress.com/2012/09/06/convert-bzr-to-git/) to convert the Launchpad bazaar repository to git:

I then followed [this explanation on StackOverflow](http://stackoverflow.com/questions/1220557/how-do-i-prepend-history-to-a-git-repo/1547490#1547490) to create the link using Git replace using the method described in the section *"Using refs/replace/ mechanism"*.

Ede123:
[...]

PR #249, opened 2020-05-21
Optimize out some redundant code or calls to "slow" functions

nthykier:


scour/scour.py
```diff
@@ -3627,8 +3625,8 @@ def scourString(in_string, options=None):
     # on the first pass, so we do it multiple times
     # does it have to do with removal of children affecting the childlist?
     if options.keep_editor_data is False:
-        while removeNamespacedElements(doc.documentElement, unwanted_ns) > 0:
-            pass
+        _num_elements_removed += removeNamespacedElements(doc.documentElement,
```
Ede123, line 3631: Would be great if this were true, but someone added the comment above for some reason at some point, so I wonder if the loop can really be removed?

scour/scour.py
```diff
@@ -3631,8 +3629,8 @@ def scourString(in_string, options=None):
     if options.keep_editor_data is False:
         while removeNamespacedElements(doc.documentElement, unwanted_ns) > 0:
             pass
-        while removeNamespacedAttributes(doc.documentElement, unwanted_ns) > 0:
-            pass
+        _num_attributes_removed += removeNamespacedAttributes(doc.documentElement,
```
Ede123, line 3633: dito

scour/scour.py
nthykier: Honestly, I cannot see how that comment can still be true.  Unfortunately, both the functions and the comments predates the version history in git, so it is not possible to tell if it underlying problem was fixed and the comments were just left there to rot.

But if the comments are true, then we are missing a regression test case to show the issue.  Sadly, it is a bit of a devil's proof and my head is convinced that there is no such case (doesn't make it true; I cannot find a case but my thinking is biased).

I am fine with ripping those two commits out of the branch if you are not comfortable with them.

scour/scour.py
nthykier: Ditto as well :)

scour/scour.py
```diff
@@ -74,6 +74,9 @@
 COPYRIGHT = u'Copyright Jeff Schiller, Louis Simard, 2010'
 
 
+# Used to split values where "x y" or "x,y" or a mix of the two is allowed
+SPACE_COMMA_SEPARATOR = re.compile(r"\s*[\s,]\s*")
```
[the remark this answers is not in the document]
nthykier: Done

scour/scour.py
Ede123: I think you're right, let's keep the commits and remove the comment!

Note you can [get full history](https://github.com/scour-project/scour/issues/23#issuecomment-241896530) with `git fetch origin 'refs/replace/*:refs/replace/*'` or via [old bzr history](https://bazaar.launchpad.net/~scouring/scour/trunk/changes) on Launchpad.

Looking at that, the comment was added in https://github.com/scour-project/scour/commit/879300373f39319c76ef58bd10ce0e32ee55d22b and fixed already in https://github.com/scour-project/scour/commit/2dc788aa3f5196a14c00262238ba838248a48077.

PR #257, opened 2020-06-27
Fix #256 - renameID bug plus too few IDs allocatable for remapping

### LINKED ISSUE #256: --shorten-ids breaks file and increases file-size
`scour -i input -o output --shorten-ids` breaks file

## Input
[Input.svg](https://gitlab.com/inkscape/inbox/uploads/39f2912ed29ffeb15111f0318672ba2d/Input.svg)

## Output
[Output.svg](https://gitlab.com/inkscape/inbox/uploads/7b1de555ac8664fc65adcc1de7261810/Output.svg)

## Changes
only ids are changing

The file contains more than 26 references and uses (a-z and A-Z), however scour changes to (a-z and aa-az)

I think there might be somehow an overflow-error in the ids?

PS the files contains **no** CSS

## Source
https://commons.wikimedia.org/wiki/File:Bundesrat_Sitzverteilung.svg

nthykier:
This PR solves #256, which is really two issues combined:

1. There was a bug in `renameID` where if two IDs were swapped (`AA <=> a`), then all references would end up in one of the IDs (rather than correctly split between the two).
1.  scour would only allocate 26 characters (`a-z`) for IDs.  This branch bumps it to 52 (`a-zA-Z`), which solves the size issue in #256.  Though in theory it can be increased even further (e.g. `0-9` can be used from character 2 and later in the ID) - this is left as future work.

Closes: #256

### DIFF
--- a/scour/scour.py
+++ b/scour/scour.py
@@ -847,8 +847,6 @@ def renameID(idFrom, idTo, identifiedElements, referringNodes):
 
     definingNode = identifiedElements[idFrom]
     definingNode.setAttribute("id", idTo)
-    del identifiedElements[idFrom]
-    identifiedElements[idTo] = definingNode
     num += len(idFrom) - len(idTo)
 
     # Update references to renamed node
--- a/scour/scour.py
+++ b/scour/scour.py
@@ -786,20 +786,23 @@ def shortenIDs(doc, prefix, options):
     return num
 
 
+ID_CHARS = [chr(i + ord('a')) for i in range(26)] + [chr(i + ord('A')) for i in range(26)]
+
+
 def compute_id_lengths(highest):
     """Compute how many IDs are available of a given size
 
     Example:
         >>> lengths = list(compute_id_lengths(512))
         >>> lengths
-        [(1, 26), (2, 676)]
+        [(1, 52), (2, 2704)]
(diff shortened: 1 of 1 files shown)


nthykier:
@Ede123 Did you have a chance to review this?

Ede123:
Sorry, my availability varies these days and I'm usually busy on the weekends. I'll try to find some time to review  (hopefully soon).

Ede123:
Thanks for this, I've just merged the bug fix in the first commit in f0788d5c0d49ee93e2aad44654cbc4ab22ffa70b .

I have to think a bit about the second commit (Support uppercase letters in generated IDs) - while it's obviously true we can save some bytes by using upper case characters it adds ambiguity ("a" vs "A" might be a different code point but it's still the same letter). Also, as you mention yourself, if we wanted to maximize the number of possible IDs we could allow a lot more characters and squeeze even more bytes out of it (at the cost of legibility).

Feel free to open a new issue about this so we can discuss a bit, if you think optimizing ID generation is a worthwhile task.

Ede123:
Apparently I have to comment to close this :facepalm:

Ede123:
a

Ede123:
Sorry for this, something's broken...

Issue #259, opened 2020-08-02
The --set-precision option is confusing


mrmeszaros:
Hey,
As I understand `precision` controls the number of significant digits.

However, this means that close to zero, the grid is finer, while farther away it gets coarser. For example:

```
<svg>
  <path d="m 100.501 .999"/>
</svg>
```

Now if I run it with `--set-precision=3`, it produces the following:
```
<svg>
  <path d="m101 0.999"/>
</svg>
```
I suggest to either change the behavior to round to decimal places, or to introduce a new option that implements that kind of rounding.

For example: `--set-rounding=N` will round numbers to the nearest `1e-N`. This would work like python's round `built-in` method.

Ede123:
Basically a duplicate of #141.

See comments there for some thoughts.

PR #262, opened 2020-08-14
Add option for rounding to set number of decimals

### LINKED ISSUE #141: Precision relative to viewbox or max demical point
Ok I re-thought the problem with precision. @Ede123 I think your right about the zero being arbitrary related to #136. This is not an attempt to waste your time and sound like I'm repeat myself.  Not sure if the scour project is interested in going in this direction. But thought I'd share anyway.

I think the reason that rounding near zero worked well for me was not that 0 is special. Its that the precision operation doesn't account for the dimensions of the svg image viewbox. 

If I have a viewbox 100px width 100px height, normally I'm not too concerned with precision like 0.00001. Actually it seems the smallest number i work with is intervals of 0.01.  So 100/0.01 =  10000 units. Which looks like 6 significant digits.

Similarly if i have a viewbox 30px by 30px, using the same 10000 units, my smallest signficant interval would be 0.003. 

So how does a number like 0.00005 slip by scour, but not 1.00005?

The suggestion is to implement some limiting option like --max-decimal-point, or --precision-relative-to-viewbox, or even implicity limit the max to 10,000/100,000 units  the max width/height.

mrmeszaros:
Fixes #141

### DIFF
--- a/scour/scour.py
+++ b/scour/scour.py
@@ -2821,8 +2821,8 @@ def parseListOfPoints(s):
     i = 0
     while i < len(nums):
         try:
-            nums[i] = getcontext().create_decimal(nums[i])
-            nums[i + 1] = getcontext().create_decimal(nums[i + 1])
+            nums[i] = getcontext().create_decimal(nums[i]).quantize(roundingQuantum)
+            nums[i + 1] = getcontext().create_decimal(nums[i + 1]).quantize(roundingQuantum)
         except InvalidOperation:  # one of the lengths had a unit or is an invalid number
             return []
 
@@ -3641,8 +3641,10 @@ def scourString(in_string, options=None):
     # to minimize errors
     global scouringContext
     global scouringContextC  # even more reduced precision for control points
+    global roundingQuantum
     scouringContext = Context(prec=options.digits)
     scouringContextC = Context(prec=options.cdigits)
+    roundingQuantum = Decimal(10) ** (-options.rounding)
 
     # globals for tracking statistics
     # TODO: get rid of these globals...
@@ -3976,6 +3978,9 @@ _option_group_optimization.add_option("--set-c-precision",
                                       action="store", type=int, dest="cdigits", default=-1, metavar="NUM",
                                       help="set number of significant digits for control points "
                                            "(default: same as '--set-precision')")
+_option_group_optimization.add_option("--set-rounding",
+                                      action="store", type=int, dest="rounding", default=9, metavar="NUM",
+                                      help="set number of digits after decimal point (default: %default)")
 _option_group_optimization.add_option("--disable-simplify-colors",
                                       action="store_false", dest="simple_colors", default=True,
                                       help="won't convert colors to #RRGGBB format")


Ede123:
I'm still mulling over this. In general I'm not opposed, but it has the potential to break images if used wrongly (which the existing precision option usually does not!) and on top of that it will be confusing for people to have two separate options to control numeric precision.

I might consider accepting it as an "interim solution" until we can come up with something that ticks all checkboxes, i.e. is user-friendly, ensures image quality *and* achieves smallest file size.

A "proper" fix likely would be what I suggested in https://github.com/scour-project/scour/issues/141#issuecomment-282844930, as I ensures precision matches the size of objects, but it will be a bit more work to implement (also as absolute coordinates would probably need to be treated differently from relative coordinates).

In any case the MR can't be accepted as-is, as it's currently rounding the numbers on *import*. That means numeric errors will be "amplified" while processing the file (and doing things like applying transforms, etc.). Instead any rounding needs to be done at the very end of all processing when writing out the numbers as a string.

mrmeszaros:
Hey, thanks for the review!

I totally agree, rounding / precision should be done on output, and not on import.

I will try to look into how it can be implemented. If you have any ideas, please share :)

codeling:
I would really like such an "absolute precision" mode (i.e. being able to determine how many digits after the comma the result has). For me it makes more sense than the current mode, which I would call "relative precision".

For positions in the current, "relative" precision mode it depends where in the image a coordinate is to determine its actual precision in the output. Let me give an example of what I mean:
- "precision" set to 3
- some object a at position 100.123, 100.123 -> shortened to 100, 100 (meaning also that some object b at 100.4, 100.4 would be "shifted" to be at exactly the same position as a)
- another box at position 0.123, 0.123 -> remains at 0.123, 0.123 (and in contrast to the  case above, an object at 0.4, 0.4, so with the same offset as b has to a above, would remain at its distinct position and not be "shifted" to the same position)
So in the current mode, positions closer to the origin (with less digits before the comma) are stored with much higher overall precision in the output!

As this examples below show, this has the potential to introduce different look of regions in the resulting images **depending on where in the image objects are positioned!**.

Example .svg file:
```
<svg version="1.1" viewBox="0 0 101 101" xmlns="http://www.w3.org/2000/svg">
 <g fill="#444">
  <rect x=".123" y=".123" width=".2" height=".2"/>
  <rect x=".4" y=".4" width=".2" height=".2"/>
  <rect x="100.123" y="100.123" width=".2" height=".2"/>
  <rect x="100.4" y="100.4" width=".2" height=".2"/>
 </g>
</svg>
```

Output (produced using Inkscape 1.2.2 with Scour "0.31+", as far as I understand the settings regarding precision haven't changed since?):
```
<svg version="1.1" viewBox="0 0 101 101" xmlns="http://www.w3.org/2000/svg">
 <g fill="#444">
  <rect x=".123" y=".123" width=".2" height=".2"/>
  <rect x=".4" y=".4" width=".2" height=".2"/>
  <rect x="100" y="100" width=".2" height=".2"/>
  <rect x="100" y="100" width=".2" height=".2"/>
 </g>
</svg>
```

As you can see, the two rects close to the origin are really well-preserved; but the two rects at 100.x are moved over one another!

In contrast, if there was an option for a fixed number of decimal points, the output would look like this (for e.g. 1 digits after comma):
```
<svg version="1.1" viewBox="0 0 101 101" xmlns="http://www.w3.org/2000/svg">
 <g fill="#444">
  <rect x=".1" y=".1" width=".2" height=".2"/>
  <rect x=".4" y=".4" width=".2" height=".2"/>
  <rect x="100.1" y="100.1" width=".2" height=".2"/>
  <rect x="100.4" y="100.4" width=".2" height=".2"/>
 </g>
</svg>
```
Which would use a little more space than above of course, but would be much more consistent: The distance between to rect 1 and 2 would be the same as the distance between rect 3 and 4, instead of 3 and 4 being moved to exactly the same coordinates!

Where it could make sense in my opinion to use a *"relative precision"* (i.e. significant digits overall) is for encoding length; but even then the "environment" has to be taken into account; for example:
 - If you show two bars, one of length 100.1 and 110.2, then the .1 and .2 don't matter in how the relative lengths of these two bars are perceived by viewers, even more so if they are placed apart or are not starting at the same position in the direction in which they have these lengths.
 - If in contrast they are again of length 100.1 and 100.2, are starting at the same position, and are positioned right next to one another, then viewers can just notice this slight difference, and removing it would cause the "meaning" of the figure to change.

EDIT: Just saw that somebody else has made the same point [over in that issue](https://github.com/scour-project/scour/issues/259) already!

Issue #260, opened 2020-08-07
Traceback: xml.dom.NotFoundErr with --enable-id-stripping (0.38 regression)


vbraun:
This appears to be a 0.38 regression, worked before:
~~~~
$ scour --version
0.38
$ scour --enable-id-stripping -i instagram_icon.svg 
Traceback (most recent call last):
  File "/tools/bin/scour", line 8, in <module>
    sys.exit(run())
  File "/tools/lib/python2.7/site-packages/scour/scour.py", line 4217, in run
    start(options, input, output)
  File "/tools/lib/python2.7/site-packages/scour/scour.py", line 4185, in start
    out_string = scourString(in_string, options).encode("UTF-8")
  File "/tools/lib/python2.7/site-packages/scour/scour.py", line 3823, in scourString
    while removeNestedGroups(doc.documentElement) > 0:
  File "/tools/lib/python2.7/site-packages/scour/scour.py", line 1050, in removeNestedGroups
    g.parentNode.insertBefore(g.firstChild, g)
  File "/tools/lib/python2.7/xml/dom/minidom.py", line 90, in insertBefore
    newChild.parentNode.removeChild(newChild)
  File "/tools/lib/python2.7/xml/dom/minidom.py", line 165, in removeChild
    raise xml.dom.NotFoundErr()
xml.dom.NotFoundErr
~~~~
The svg file in question is attached (rename instagram_icon.svg.txt -> instagram_icon.svg since github does not allow svg attachments)

[instagram_icon.svg.txt](https://github.com/scour-project/scour/files/5040471/instagram_icon.svg.txt)

Ede123:
Caused by dd2155e5764d13e03ee3702fdf78d48628a84988. @nthykier can you have a look?

nthykier:
Thanks for the notification; I have created #261 which fixes the crash for me.  I have not had time to understand how to create a minimal test case for the issue. However, I figured a published fixed with no test was better than nothing at all.

MattiaFailla:
Hi guys, I also ran into this problem. 
If you want I can write some CI to verify that the library behaves as it should, what do you think?

Ede123:
@MattiaFailla We already have a good amount of tests and automatic testing via Travis CI. New testcases for increased test coverage are always welcome!

(P.S. I hope to be able to commit the fix next week; I'm on holiday this week and mostly AFK)

pbregener:
[...]

PR #264, opened 2020-09-18
scour: Fix another variant of the crash from #260

nthykier:


dmedvinsky:
Any plans on merging this?

Ede123:
Thanks for the fix Nils and sorry for the delayed review!

I believe the important thing here is to avoid attempting to modify the list directly and remember to use `xml.dom`, methods like `appendChild` or `removeChild`.

With this in mind, I assume we can avoid cloning said nodes only to discard the originals later. A simple `appendChild` should be enough (it even removes the node from the original parent for us as as [per documentation](https://docs.python.org/3/library/xml.dom.html#xml.dom.Node.appendChild))

Would you be willing to give this approach another try before I merge this @nthykier ?

nthykier:
@Ede123: Rebased.

Ede123:
Thanks a lot!

PR #266, opened 2020-12-11
Add support to ppc64le. Disable pypy as this is not supported

gururajrkatti:
This PR adds CI support for the IBM Power Little Endian (ppc64le) architecture. The idea is to ensure that the builds on this architecture are continuously tested along with the Intel builds (amd64) as this is part of the ubuntu distro on that architecture as well and detecting (and fixing) any issues 
or failures early would help to ensure that we are always up to date.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,10 @@ sudo: false
 
 language: python
 
+arch:
+  - amd64
+  - ppc64le
+
 python:
   - pypy
   - 2.7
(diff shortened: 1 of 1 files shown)


gururajrkatti:
Hi just wanted to check if you got a chance to look into this.

Ede123:
Thanks for this and sorry for slow response!

Can you clarify a bit how testing on ppc64le will profit Scour? Naively I'd assume Python to work the same irregardless of the target architecture (unless there's a bug in Python, but I hope that they in turn have suitable CI to catch this precise issue).

If there's a justified risk of some scour-specific functionality to depend heavily on target architecture I'd be happy do accept this, but if it's just "testing for testing's sake" I'd rather avoid the added complexity.

gururajrkatti:
Hi @Ede123 Thanks for the response. I understand that pure python package should work fine on other platforms. However based on our experience earlier, even python packages can fail on other platforms because of an indirect dependency on native code or libraries via other python dependencies. Hence this is proposed to catch any such issues early. Tagging @gerrith3 for any additional info.

PR #267, opened 2020-12-20
Add style argument which can convert SVG attributes to CSS Style attribute (the opposite of style_to_xml.)

doom-goober:
Currently, scour allows the user two options when it comes to SVG Attributes versus CSS Style: They can either convert everything to SVG Attributes or leave all the attributes as they were originally encoded. Obviously, there is a third option, which is currently missing: convert everything to CSS Style.

While SVG Attributes are generally preferred for various CSS overriding reasons, some tools don't handle them well and prefer CSS Style. For example, Inkscape's "Save As HTML 5 Canvas" doesn't support SVG Attributes but does support CSS Style.  For this reason, tools like Illustrator allow the user to choose between "Inline" or "Presentational Attributes" (or a third option, "Internal CSS" which is out of scope for this discussion.) The eventual goal of this change would be to add an option to Inkscape so it behaves somewhat like Illustrator with three options: "Preserve", "Presentational Attributes", "Inline".

This pull request adds the option to convert styling all to CSS Style through --style=inline. The other options are --style=none, --style=attributes, --style=preserve. 

The backwards compatibility logic is slightly annoying. If you specify a --style other than "none", then that style will be used. If you specify --style=none (or specify nothing, as none is the default) then --disable-style-to-xml or --style-to-xml will decide the behavior. Perhaps "none" could renamed "default?"

### DIFF
--- a/scour/scour.py
+++ b/scour/scour.py
@@ -1819,7 +1819,7 @@ def repairStyle(node, options):
                 if propName in svgAttributes:
                     node.setAttribute(propName, styleMap[propName])
                     del styleMap[propName]
-      
+
         _setStyle(node, styleMap)
 
     # recurse for our child elements
@@ -3999,7 +3999,7 @@ _option_group_optimization.add_option("--disable-style-to-xml",
                                       help="won't convert styles into XML attributes")
 _option_group_optimization.add_option("--style",
                                       action="store", type="string", dest="style_type", default="none", metavar="TYPE",
-                                      help="style type (overrides style-to-xml): none, preserve, inline,"\
+                                      help="style type (overrides style-to-xml): none, preserve, inline,"
                                       "attributes (default: %none)")
 _option_group_optimization.add_option("--disable-group-collapsing",
                                       action="store_false", dest="group_collapse", default=True,
--- a/scour/scour.py
+++ b/scour/scour.py
@@ -1805,7 +1805,7 @@ def repairStyle(node, options):
         elif options.style_type == "preserve":
             # Keep whatever style of attribute versus style the file currently has
             pass
-        elif options.style_to_xml or options.style_type == "attributes":
+        elif options.style_type == "attributes":
             # now if any of the properties match known SVG attributes we prefer attributes
             # over style so emit them and remove them from the style map
             for propName in list(styleMap):
@@ -4083,6 +4083,10 @@ def parse_args(args=None, ignore_additional_args=False):
         _options_parser.error("Input filename is the same as output filename")
     if options.style_type not in ['none', 'preserve', 'attributes', 'inline-css']:
         _options_parser.error("Invalid value for --style, see --help")
+
+    #For backwards compatibility, we support style_to_xml but style_type can override it.
+    if options.style_type == 'none' and options.style_to_xml:
+      options.style_type = 'attributes'
 
     return options
 
--- a/scour/scour.py
+++ b/scour/scour.py
@@ -1800,15 +1800,27 @@ def repairStyle(node, options):
                 del styleMap['overflow']
                 num += 1
 
+    if node.nodeType == Node.ELEMENT_NODE:
+      if options.style_type == "inline":
+        # Prefer inline style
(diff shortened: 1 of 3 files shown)


scour/scour.py
```diff
@@ -1800,9 +1800,21 @@ def repairStyle(node, options):
...
-        if options.style_to_xml:
+    if node.nodeType == Node.ELEMENT_NODE:
+        if options.style_type == "inline":
+            # Prefer inline style
+            # Remove known SVG attributes and store their values in style attribute
+            attributes = [node.attributes.item(i).nodeName for i in range(node.attributes.length)]
+            for attribute in attributes:
+                if attribute in svgAttributes:
+                    styleMap[attribute] = node.getAttribute(attribute)
+                    node.removeAttribute(attribute)
```
nthykier, line 1811: Drive by-remark, I think the following will be slightly better as it avoids another attribute lookup.

```suggestion
            attributes = [node.attributes.item(i) for i in range(node.attributes.length)]
            for attribute in attributes:
                name = attribute.nodeName
                if name in svgAttributes:
                    styleMap[name] = attribute.nodeValue
                    node.removeAttribute(name)
```

(note: hand written directly in github, it may contain typoes).

Otherwise, this part LGTM.

scour/scour.py
```diff
@@ -3985,6 +3997,10 @@ def format_usage(self, usage):
 _option_group_optimization.add_option("--disable-style-to-xml",
                                       action="store_false", dest="style_to_xml", default=True,
                                       help="won't convert styles into XML attributes")
+_option_group_optimization.add_option("--style",
+                                      action="store", type="string", dest="style_type", default="none", metavar="TYPE",
+                                      help="style type (overrides style-to-xml): none, preserve, inline,"
```
nthykier, line 4002: Should we call it `inline-css` rather than `inline` to be more explicit?  At least, I did not connect `inline` with an inline CSS attribute at first.

scour/scour.py
```diff
@@ -1800,9 +1800,21 @@ def repairStyle(node, options):
...
+            # Remove known SVG attributes and store their values in style attribute
+            attributes = [node.attributes.item(i).nodeName for i in range(node.attributes.length)]
+            for attribute in attributes:
+                if attribute in svgAttributes:
+                    styleMap[attribute] = node.getAttribute(attribute)
+                    node.removeAttribute(attribute)
+        elif options.style_type == "preserve":
+            # Keep whatever style of attribute versus style the file currently has
+            pass
+        elif options.style_to_xml or options.style_type == "attributes":
```
nthykier, line 1815: I think it would make better sense to remove `options.style_to_xml` from here and have the initial option parsing do the relevant shuffling of values.

nthykier (commented):
Hi,

Looks good from my PoV.  I have added some suggestions as a occasional contributor.

doom-goober:
@nthykier Sorry for leaving this pull request open for so long. Real life threw some turbulence at me, but I have finally incorporated your suggestions. Sorry again.

nthykier:
Hi @doom-goober

No worries. I am not a member of the scour project, so my words were only intended as suggestions and nothing more. :)

joneuhauser:
Copying over my comment from the downstream issue in Inkscape:

It's not that simple, unfortunately, because inline CSS and presentation attributes have not the same specificity (inline CSS: specificity infinity, presentation attributes: [specificity 0](https://www.w3.org/TR/SVG/styling.html#PresentationAttributes)).

In practice that means: Try adding style="fill:blue;" or fill="blue" to the element in Gitlab op's [source.svg](https://gitlab.com/inkscape/extensions/uploads/9a0edcb8f92018711405bf61a9c8e043/source.svg). For the former, the element becomes blue, for the second, it doesn't. As soon as there is any stylesheet affecting the element, even a path {} selector, it will take precedence over presentation attributes.

Just changing inline css to presentation attributes or vice versa is not going to cut it (that is also true for the existing code converting CSS to presentation attributes). I wouldn't want my SVG optimizer to change the way the exported image is rendered.

PR #268, opened 2021-01-10
Refactor bits

nthykier:
Largest bit is removal of global variables but also includes a few minor other things.

### DIFF
--- a/scour/scour.py
+++ b/scour/scour.py
@@ -1381,12 +1381,11 @@ def removeUnusedAttributesOnParent(elem):
             unusedAttrs[attr.nodeName] = attr.nodeValue
 
     # for each child, if at least one child inherits the parent's attribute, then remove
-    for childNum in range(len(childElements)):
-        child = childElements[childNum]
+    for child in childElements:
         inheritedAttrs = []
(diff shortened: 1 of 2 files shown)


scour/scour.py
```diff
@@ -1022,6 +1022,15 @@ def removeDescriptiveElements(doc, options):
     return num
 
 
+def g_tag_is_unmergeable(node):
```
Ede123, line 1019: Only very minor suggestion: Would you mind rewriting this as `g_tag_is_mergeable` instead? Avoids the double negation (`if not g_tag_is_unmergeable(child)`) below.

Ede123 (commented):
Hi Nils, this is great!

Stripping out the stats globals was long overdue :-).

scour/scour.py
```diff
@@ -2134,14 +2134,17 @@ def removeDefaultAttributeValue(node, attribute):
     return 0
 
 
-def removeDefaultAttributeValues(node, options, tainted=set()):
+def removeDefaultAttributeValues(node, options, tainted=None):
```
Ede123, line 2137: Nice, you taught me something today!

PR #270, opened 2021-02-23
Rewrite colors into shortest possible name

nthykier:
Improve the code for rewriting colors into recognising that some
colors are shorter by name.  This commit enables scour to rewrite
`rgb(255, 0, 0)` into `red` which is slightly shorter than `#f00`
(ditto for `tan` and `pink`).

When the color name ties in length with the hexcode, then scour will
leave it as-is if the input file used a variant of same length
(e.g. `blue`, `cyan` and `aqua` will be left as-is).  But if scour is
rewriting the color code, it will prefer the hex code variant.

### DIFF
--- a/scour/scour.py
+++ b/scour/scour.py
@@ -2245,6 +2245,9 @@ def convertColor(s):
     """
        Converts the input color string and returns it to the shortest known format for it.
     """
+    if s[0] == '#':
+        # Normalize #RGB into lower case early as our lookup relies on the lower case name
+        s = s.lower()
     color = colors.get(s)
     # Short cut: if we know the color (either by name or hex code) then skip the
     # parsing logic.  This makes
@@ -2255,7 +2258,6 @@ def convertColor(s):
         # If it is in #RGB/#RRGGBB format already, then we can also avoid
         # the regex parsing code.  Notably, if it is not a known hex code,
         # then at best we can truncate #RRGGBB into #RGB.
-        s = s.lower()
         if len(s) == 7 and s[1] == s[2] and s[3] == s[4] and s[5] == s[6]:
             s = '#' + s[1] + s[3] + s[5]
         return s
--- a/scour/scour.py
+++ b/scour/scour.py
@@ -4043,7 +4043,7 @@ _option_group_optimization.add_option("--set-c-precision",
                                       action="store", type=int, dest="cdigits", default=-1, metavar="NUM",
                                       help="set number of significant digits for control points "
                                            "(default: same as '--set-precision')")
-_option_group_optimization.add_option("--disable-simplify-colors",
+_option_group_optimization.add_option("--disable-shorten-colors", "--disable-simplify-colors",
                                       action="store_false", dest="simple_colors", default=True,
                                       help="won't convert colors to a shorter format format")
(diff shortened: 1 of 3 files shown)


Ede123:
Would this still create a group and propagate the `fill` to the parent for the following?
```
<rect fill="#f00" x="10" y="10" width="10" height="10"/>
<rect fill="red" x="20" y="20" width="10" height="10"/>
```

nthykier:
> Would this still create a group and propagate the `fill` to the parent for the following?
> 
> [...]

Neither the old nor the new code gets this consistently right because that is governed by another part of the code. For reference, the new code gets your particular example correctly. However, I would expect both the old and the new code to fail that test with e.g. `cyan` vs `aqua` vs `#0ff`.

For that to (always) work, then we would also have to change:

```
            newColorValue = convertColor(oldColorValue)
            oldBytes = len(oldColorValue)
            newBytes = len(newColorValue)
            if oldBytes > newBytes:
                element.setAttribute(attr, newColorValue)
                numBytes += (oldBytes - len(element.getAttribute(attr)))
```

Notably, the `if oldBytes > newBytes:` which should be rewritten as `if oldBytes >= newBytes and oldColorValue != newColorValue:`

Ede123:
Yep, you're right, as long as we convert consistently (even if it's not shorter but the same length) we should be fine optimization-wise. We should probably add a few testcases for these border cases as well...

Remaining questions I'm mulling about:
* The related preference is named `--disable-simplify-colors`. You already adjusted the description, but I figure the name is not really suitable anymore. Maybe we should consider renaming it (keeping the old name as fall-back as we did for other preferences), e.g. `--disable-shorten-colors`?
* I have to admit seeing RGB colors and color names mixed in a document makes my skin crawl. Is this just me? Objectively speaking going for the shortest name seems reasonable, so I probably should bite the bullet here, but I wanted to put it out there for consideration at least. 😉

nthykier:
Ok, I have:

 * Added `--disable-shorten-colors` as an alias of `--disable-simplify-colors`
 * Made scour always normalize to the _same_ variant of a color (except when we are not shorting colors).
 * Fixed a bug in handling of upper case hex code writing two scour runs to be normalized to the shortest possible code.

nthykier:
As for mixing color codes with color names - scour is here to write short svg files, not beautiful ones. :)

Also, if your beef with this is about consistency, then we are trying one consistency (form) for another (shortest length). :smile:

scour/scour.py
```diff
@@ -2194,40 +2241,59 @@ def removeDefaultAttributeValues(node, options, tainted=set()):
...
-       Converts the input color string and returns a #RRGGBB (or #RGB if possible) string
+       Converts the input color string and returns it to the shortest known format for it.
     """
-    s = value
+    if s[0] == '#':
+        # Normalize #RGB into lower case early as our lookup relies on the lower case name
+        s = s.lower()
+    color = colors.get(s)
+    # Short cut: if we know the color (either by name or hex code) then skip the
+    # parsing logic.  This makes
```
eweitz, line 2253: This looks like an incomplete thought.

eweitz:
As someone who independently developed a rough version of this, I'd find this functionality useful.

Could a Scour maintainer chime in to note what could move this PR forward?

PR #275, opened 2021-04-06
Fix #274 - Overwrite svg file if input file is output

### LINKED ISSUE #274: Overwrite svg file if input file is output
Hi, if input and output are the same file, there will be an error `Input filename is the same as output filename`.
But I think, that this is valid case of usage.

gistrec:
This PR solves #274

### DIFF
--- a/scour/scour.py
+++ b/scour/scour.py
@@ -3970,6 +3970,10 @@ _option_group_optimization.add_option("--no-renderer-workaround",
 _options_parser.add_option_group(_option_group_optimization)
 
 _option_group_document = optparse.OptionGroup(_options_parser, "SVG document")
+
+_option_group_document.add_option("--allow-file-overwrite",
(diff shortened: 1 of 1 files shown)


Ede123:
Thanks for this!

Some thoughts
* I'm still thinking about border cases that might destroy the original file content when something goes wrong - we should make absolutely sure this can not happen. My first thought was to write to a temporary file first, and move/rename later, but that might have implications of it's own and might mean unnecessary file IO.
* I find the approach of opening one file in `r+` mode and still treating it as two separate files (input/output) a bit convoluted and see the danger of introducing bugs at one point if we're not careful. That said I admit it is an elegant solution in principle, but maybe we can still come up with something better, that makes it clear it's only one file or actually handles reading and writing separately?
* I'd prefer the option name `--in-place` in line with the original issue https://github.com/scour-project/scour/issues/129

PR #280, opened 2021-06-05
Scour animated paths

yttiktak:
animate/attributeName/d needs special treatment to keep node counts equal

### DIFF
--- a/scour/scour.py
+++ b/scour/scour.py
@@ -2766,6 +2766,18 @@ def clean_path(element, options, stats):
         stats.num_bytes_saved_in_path_data += (len(oldPathStr) - len(newPathStr))
         element.setAttribute('d', newPathStr)
 
+def clean_animated_path(element, options, stats):
+    """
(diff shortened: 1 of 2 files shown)


scour/scour.py
```diff
@@ -2766,6 +2766,18 @@ def clean_path(element, options, stats):
         stats.num_bytes_saved_in_path_data += (len(oldPathStr) - len(newPathStr))
         element.setAttribute('d', newPathStr)
 
+def clean_animated_path(element, options, stats):
+    """
+    Cleans (precision only) the path strings (values that will replace a path's d attribute) of the animation element
+    These are ';' delimited path data that must keep the same number of nodes, so the only cleaning is via serializePath   
+    """
+    oldPathStr = element.getAttribute('values')
+    oldPathStrs = oldPathStr.split(';')
```
nthykier, line 2775:
```suggestion
    if oldPathStr == '':
        return
    oldPathStrs = oldPathStr.split(';')
```

